PR #6, opened 2018-10-12
Moved out ruby

Albinutte:
Moved Ruby out! To test, I suggest running some Ruby code on any algorithmic task, and then try solving [migrateDatabase](https://github.com/CodeSignal/content/tree/master/_devops_tasks/migrateDatabase); pasting code from `solution_migrateDatabase.rb` should pass all test cases.


ruby/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM codesignal/ubuntu-base:v1.0
```
ksaxberg: Live is bad and full of terrors, you'll need to increment this version to target the yet-to-be-created updatedbase dockerfile. Maybe `v1.1`?

Albinutte: Oh, no! Will change that.

PR #7, opened 2018-10-12
Added java dockerfile

Albinutte:


java/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM codesignal/ubuntu-base:v1.0
+
+RUN add-apt-repository -y ppa:linuxuprising/java \
+ && apt-get update \
+ && echo oracle-java10-installer shared/accepted-oracle-license-v1-1 select true | /usr/bin/debconf-set-selections \
+ && apt-get install -y --no-install-recommends oracle-java10-installer oracle-java10-set-default libjackson2-core-java libjackson2-databind-java libmysql-java \
```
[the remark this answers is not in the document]
ksaxberg: Also, should we put `maven` in here?

java/Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+ && rm -rf /var/lib/apt/lists/*
+
```
ksaxberg: At the [very bottom of the main dockerfile](https://github.com/CodeSignal/dockerfiles/blob/f9d7e158db75bfc906fd758eb0546070a6fdd6f9/main/Dockerfile#L273), it looks like there are a few additional java dependencies for working with sql. I don't think all of this is relevant to java, but at least a few of these are necessary. 

It's going to be great once we split this out!

Albinutte: I did add one that had `java` in its name; do you think something else is missing?

ksaxberg: Oh! That was the only one I thought might be useful. My eyes aren't what they used to be, I missed that you had already included it! Nice!

java/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM codesignal/ubuntu-base:v1.0
+
+RUN add-apt-repository -y ppa:linuxuprising/java \
```
ksaxberg: Not sure if this `ppa` is accurate? The build failed at this step, rejecting the package. I'll try restarting the docker service and re-running. It might be cause to re build the base, if the team is a newer one and our base image needs a refreshed list. (Or are we scrubbing the lists every time? I think apt-repository might be using the same directory that we're clearing after every run with `rm -rf /var/lib/apt/lists/*`)

screen shot 2018-10-15 at 3 03 02 pm

Albinutte: I had the same issue until I restarted docker server 😕

ksaxberg: Yup, restarting the docker server worked 👍

ksaxberg:
@Albinutte Oh yea, maybe update the version to base `1.1`

PR #15, opened 2018-10-19
C++ and Swift upgrades

### LINKED ISSUE #5: Have C++ Dockerfile extend base
It would be great if the C++ Dockerfile could extend our Base dockerfile. Here's a little rework to help get you started: (untested / unbuild, but passes `fromlatest.io`)

```
FROM codesignal/ubuntu-base:v1.0

RUN npm install -g \
    underscore \
    request \
    shelljs \
    http \
    sys \
    jquery \
    async \
    connect \
    validator \
    ws \
    co \
    when \
    forever \
    debug \
  && add-apt-repository -y ppa:ubuntu-toolchain-r/test \
  && apt-get update \
  && apt-get install -y --no-install-recommends \
    python-software-properties \
    curl \
    libcurl4-openssl-dev \
# Installing G++ 8.x version from the ppa
    g++-8 \
    gcc-8 \
# Install GNU Fortran 8
    gfortran-8 \
# Additional C++ dependencies
    libssl-dev \
    libjsoncpp-dev \
# Clean up

```

ksaxberg:
Fixes #5 

Upgraded both the `C++` and `Swift` container to extend the ubuntu-base. Hopefully we'll be able to upgrade the swift language soon, but I didn't want to take that on during a potentially breaking container change.

Since the `InstallObjectiveC` file is run with sudo permissions by default, I removed `sudo` from the commands here. There is potentially a breaking change, as some commands were run with `sudo -E` which preserves environment variables, but it seems to run just fine after testing so I don't think it broke anything.


swift/Dockerfile
```diff
@@ -1,52 +1,39 @@
...
     && gpg --verify "${SWIFT_ARCHIVE_SIGNATURE}" \
     && tar -zxvf "${SWIFT_ARCHIVE}" --directory / --strip-components=1 \
+# Install Objective C
+    && chmod 777 /root/InstallObjectiveC.sh \
+    && /bin/bash -c "/root/InstallObjectiveC.sh" \
+# Cleanup
+    && rm -r "${GNUPGHOME}" "${SWIFT_ARCHIVE}" "${SWIFT_ARCHIVE_SIGNATURE}" \
+    && apt-get -qq -y remove curl cmake git \
+    && apt-get -qq -y autoremove \
+    && rm -rf /var/lib/apt/lists/*
```
Albinutte: Should you also delete ` "/root/InstallObjectiveC.sh"`?

ksaxberg: Arguably yea, it's like 1kb tho

Albinutte: Ya know what they say, if you look after the kbs, the Mbs look after themselves!

Albinutte:
I'm afraid Swift / ObjC are broken 😞 If you try to run a free-coding template, you get 
```
/usercode/file.objc:3:9: fatal error: 'Foundation/Foundation.h' file not found
#import <Foundation/Foundation.h>
        ^
1 error generated.

No output
```

ksaxberg:
Added back in all the curious dependencies from the dockerfile (including rsync?!). It would be really good to either use multi-stage builds to leverage someone's pre-build swift image or to re-investigate if there are easier ways to install swift. (Hoping maybe there's a ppa out there?)

Tested swift on `add` and it works. Tried to copy over the code we send for execution to create a test, but I kept on getting "syntax errror" when directly sending the "file.swift" code or a copy of the `code` in the request..

ksaxberg:
I added another test to `objc` over [here](https://github.com/CodeSignal/coderunner/pull/134), and this passes for me when running in a container. So it looks like the dependencies are being properly included for objc. I'll go ahead and rebuild the swift container, just to be sure

Albinutte:
@ksaxberg my bad! It all looks good 👍

PR #20, opened 2018-10-22
D

ksaxberg:
Works with both `hello` test and with running `add`. 1.09GB in size, so 250MB on the base image.

### DIFF
--- a/d/Dockerfile
+++ b/d/Dockerfile
@@ -1,7 +1,7 @@
 FROM codesignal/ubuntu-base:v1.1
 
 RUN apt-get update \
-  && apt-get install -y --no-install-recommends wget \
+  && apt-get install -y --no-install-recommends wget libcurl3 \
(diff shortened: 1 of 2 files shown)


d/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM codesignal/ubuntu-base:v1.1
+
+RUN apt-get update \
+  && apt-get install -y --no-install-recommends wget libcurl3 \
+  && wget http://downloads.dlang.org/releases/2.x/2.079.1/dmd_2.079.1-0_amd64.deb \
+  && dpkg -i dmd_2.079.1-0_amd64.deb \
```
Albinutte: Should this file be deleted after the installation?

ksaxberg: Wasn't sure if it was automatically cleaned or needed to exist for linking purposes. I'll try removing it

ksaxberg:
@Albinutte Removed the .deb, still works with both `hello` and `add`! Good catch

PR #23, opened 2018-10-26
Java bump; gradle support

Albinutte:
Updates Java version to 11 in both main and Java dockerfiles, and adds Gradle to Java dockerfile.

### DIFF
--- a/java/Dockerfile
+++ b/java/Dockerfile
@@ -12,8 +12,6 @@ RUN add-apt-repository -y ppa:linuxuprising/java \
         maven \
     && rm -rf /var/lib/apt/lists/*
 
-ENV CLASSPATH=/usr/share/java/*:$CLASSPATH
-
 # install gradle
 RUN apt-get update \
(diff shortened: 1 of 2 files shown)


Albinutte:
@ksaxberg this PR also fixes Java in main dockerfile; we'll need to merge it today so that we could redeploy prod coderunners.

Albinutte:
@ksaxberg this is just Dockerfile update; I'll be using this image to create test wrapper & stuff. Will try defining ENVs in a single command 👍

ksaxberg:
You don't have to update the ENVs. It looks like there's an issue with clojure, I'm getting syntax error on the "hello" test

Albinutte:
Clojure split is good to go 👍 Let's get all these PRs merged, and update coderunner to use correct VMs for different testing tools!

Albinutte:
(as in, I don't want to touch that main `Dockerfile` pls)

PR #27, opened 2018-10-30
Split out (almost) all the other languages

### LINKED ISSUE #11: Move out more languages
Migrating issue https://github.com/CodeSignal/coderunner/issues/91 here. This issue includes moving more languages out to separate docker files.

## Languages to Support
- [x] Bash     (ubuntu-base)
- [x] C    (cpp)
- [x] Clojure
- [x] Coffeescript
- [x] C#    (csharp)
- [x] C++      (cpp)
- [x] D
- [x] Dart
- [x] Erlang
- [x] Elixir
- [x] F#   (csharp)
- [x] Fortran (cpp)
- [x] FreePascal
- [x] Go
- [x] Groovy 
- [x] Haskell    (haskell)
- [x] Java     (java)
- [x] Javascript      (ubuntu-base)
- [x] Julia
- [x] Kotlin
- [x] Lisp
- [x] Lua
- [x] Mysql     (ubuntu-base)
- [x] Nim
- [x] ObjectiveC     (swift)
- [x] Octave     (octave)
- [x] OCaml       (ocaml) 
- [x] Perl
- [x] Php        (php)   
- [x] Postgresql
- [x] Python      (python)
- [x] Python3      (python)
- [x] R       (r)
- [x] Ruby
- [x] Rust
- [x] Scala
- [x] Smalltalk       (smalltalk)  
- [x] Swift       (swift)
- [x] Tcl       (tcl)
- [x] Typescript
- [x] VB
- [x] Rspec

ksaxberg:
Fixes #11 

Splits out the remaining explicit languages (Galen requests need to be done as well? I'll open a new issue for that)
#### Languages:
- [x] dart :white_check_mark:
- [x] erlang :white_check_mark:
- [x] elixir :white_check_mark:
- [x] groovy :white_check_mark:
- [x] julia :white_check_mark:
- [x] kotlin :white_check_mark:
- [x] lisp :white_check_mark:
- [x] perl :white_check_mark:
- [x] rust (Adjusted the install so there's a new run command, provided in [this](https://github.com/CodeSignal/coderunner/pull/148) PR) :white_check_mark:
- [x] scala :white_check_mark:
- [x] vb (relies on C# container, so you'll need to rebuild that one). :white_check_mark:

### Testing
Coderunner already has an additional "add" test for all of these languages, which mirrors exactly what the full app sends to the coderunner (I just copied a successful request and put it in a test). Let me know if there are any additional modes / ways of running the languages that should also be tested.

### DIFF
--- a/cs/Dockerfile
+++ b/cs/Dockerfile
@@ -10,6 +10,7 @@ RUN apt-get update \
        mono-xsp4 \
        mono-xsp4-base \
        mono-complete \
+       mono-vbnc \
        ca-certificates-mono \
        fsharp \
     && rm -rf /var/lib/apt/lists/*
--- a/erlang/Dockerfile
+++ b/erlang/Dockerfile
@@ -6,8 +6,7 @@ RUN apt-get update \
     && dpkg -i erlang-solutions_1.0_all.deb \
     && apt-get update \
 #Install Erland and High-Performance Erlang Project compiler
-    && apt-get install -y --force-yes erlang erlang-base-hipe \
-    && apt-get install -y erlang-base-hipe \
+    && apt-get install -y --force-yes --no-install-recommends erlang erlang-base-hipe \
 # Cleanup
     && apt-get remove -qq -y wget \
     && rm erlang-solutions_1.0_all.deb \
--- a/erlang/Dockerfile
+++ b/erlang/Dockerfile
@@ -1,0 +1,15 @@
+FROM codesignal/ubuntu-base:v1.1
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends wget \
+    && wget https://packages.erlang-solutions.com/erlang-solutions_1.0_all.deb \
+    && dpkg -i erlang-solutions_1.0_all.deb \
+    && apt-get update \
+#Install Erland and High-Performance Erlang Project compiler
+    && apt-get install -y --force-yes erlang erlange-base-hipe \
+    && apt-get install -y erlang-base-hipe \
+# Cleanup
+    && apt-get remove -qq -y wget \
+    && rm erlang-solutions_1.0_all.deb \
+    && rm -rf /var/lib/apt/lists/*
+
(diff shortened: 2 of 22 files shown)


cs/Dockerfile
```diff
@@ -10,6 +10,7 @@ RUN apt-get update \
        mono-xsp4 \
        mono-xsp4-base \
        mono-complete \
+       mono-vbnc \
        ca-certificates-mono \
        fsharp \
     && rm -rf /var/lib/apt/lists/*
```
Albinutte: Just to clarify, CS tag will have to be bumped on dockerhub, correct?

ksaxberg: .... we could just rebuild it, and alias the increased github release version to the existing docker version. (I.e. rebuild v1.1, but pointing to this new dockerfile)

Albinutte: Thaaat's also an option 👍

ksaxberg: 🤷‍♂️ I'm up for either, it depends on if we need to deploy coderunners while that one rebuilds. It would be safer not to, but it should be fine.

Albinutte: I'm mostly up for the easiest solution. Publishing this image with VB support shouldn't break anything, but it would be nice to be able to tell the difference between the versions.

erlang/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM codesignal/ubuntu-base:v1.1
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends wget \
+    && wget https://packages.erlang-solutions.com/erlang-solutions_1.0_all.deb \
+    && dpkg -i erlang-solutions_1.0_all.deb \
+    && apt-get update \
+#Install Erland and High-Performance Erlang Project compiler
+    && apt-get install -y --force-yes --no-install-recommends erlang erlang-base-hipe \
```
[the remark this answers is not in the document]
ksaxberg: Removed, rebuilt, and tested. Still works without the flag, guess it was completely unnecessary!

PR #29, opened 2018-10-31
Galen container

### LINKED ISSUE #28: Split out Galen
Need to split out the Galen environment from Main, should be the last thing!

ksaxberg:
Fixes #28 

A container for Galen. Tests pass. Final size for this one is 1.6GB, which is not shared with any base containers.


Albinutte:
This works! Buut for some reason now most of the tasks we have fail the test cases 😞 I'm afraid that changing the environment slightly changes the way stuff gets rendered, and this causes the tests to fail.

I'm afraid we'll need to ask Content Team to update test cases for all the tasks if we want to be able to deploy these changes, or try to change the environment somehow to fix that (although I'm not sure how possible this is).

ksaxberg:
😨 Darn. 

Maybe there are still a few hidden dependencies changing what Xvfb is rendering? Not sure why this would cause failures for our tasks, `galen` doesn't hook in to the local `npm`, right? I was assuming it was sandboxed from the rest of the local javascript environment.

Albinutte:
No idea 😞 That would be my assumption, too. The failures are super tiny, like "2 pixels mismatch" & stuff, but they will cause all of the existing solutions to fail.

ksaxberg:
Thanks for pointing out this was failing our actual tests. Iterated back from main to a 2GB Dockerfile with largely relevant dependencies. This latest container correctly runes solutions for Game of Thrones and TicTacToe frontend tasks.

PR #35, opened 2018-11-06
Julia 1.0.1

Albinutte:
Julia 1.0.1 dockerfile. You might find it somewhat familiar 🙃


julia1/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM codesignal/ubuntu-base:v1.1
```
ksaxberg: Hey... could we add a dash so this is `julia-1`? Could we also do the same over for `swift-4`?

Albinutte: Yup!

PR #39, opened 2018-11-20
Added toolz to Python

Albinutte:
Soo Python image size is `8.341 GB`. Did my changes do that?


ksaxberg:
Wait.... is it `8GB`  after only toolz?! Its supposed to be like 50kb

Albinutte:
@ksaxberg Ikr? Maybe I'm building it wrong or something 🤔

ksaxberg:
.. Let's see what happens on dockerhub

PR #41, opened 2018-12-18
Update UbuntuBase, increment all versions

ksaxberg:
Incrementing all versions after an update to ubuntu-base.

Update to `ubuntu-base` comprises adding a command to the startup `~/.bashrc` for root, which is necessary to get mysql to start up correctly. It's weird and hacky and the `chown` modifications are already made above, but something about the `overlay2` filesystem is not letting mysql startup. We could have switched to the `aufs` filesystem for docker which would also fix the issue, but that has potential performance implications.

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -55,3 +55,7 @@ RUN echo "deb http://cn.archive.ubuntu.com/ubuntu/ xenial main restricted univer
 # Clean up
     && apt-get autoremove \
     && rm -rf /var/lib/apt/lists/*
+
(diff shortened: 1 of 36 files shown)


ksaxberg:
## v2.0 Release Tracking on Dockerhub:
#### Step 1
- [x] base
#### Step 2
- [ ] java
- [ ] erlang
- [ ] rust
- [ ] clojure
- [ ] coffeescript
- [ ] cpp
- [ ] cs
- [ ] d
- [ ] dart
- [ ] go

#### Step 3
- [ ] elixir
- [ ] groovy

PR #45, opened 2019-01-15
Add curl and wget to base dockerfile

### LINKED ISSUE #44: Add Curl and Wget to base image
Add those utils in so devs have access from every code environment.

ksaxberg:
Fixes #44 

Also needed to remove all the places where we cleaned those up. Went ahead and updated all the versions as well

Didn't touch the main dockerfile, since we're planning on removing that.... eventually

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -11,6 +11,8 @@ RUN apt-get clean \
       locales \
       software-properties-common \
       build-essential \
+      curl \
+      wget \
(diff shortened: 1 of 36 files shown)


ksaxberg:
I'm just going with major version upgrades if the base changes, since that requires everything to be rebuilt. There probably should have been another number, so v0.4.0 instead of v4.0, but it's a little late to walk back now.

PR #59, opened 2019-05-31
Added PHPUnit

Albinutte:
Part of https://github.com/CodeSignal/coderunner/issues/254. This PR updates PHP docker image with PHPUnit and `php7.2-xml`, required to make it possible to get XML reports from  PHPUnit.

__To test this__

I deployed coderunner at with IP `192.241.250.45`, where I installed this new image. If you SSH into it, you will find this new `Dockerfile` in the `/coderunner` directory. I've installed it already, but you can rebuild it with `docker build -t codesignal/php:v4.1 .`.

There's also a `phpExample` directory under `/coderunner` directory. You can try running PHPUnit by building a container with it. Run the following command to start a container and get into it: `docker run -itv "/root/coderunner/phpExample":/usercode  codesignal/php:v4.1 /bin/bash`. After that `cd` into `usercode`, and run `phpunit tests --log-junit ./out.xml --no-logging`. You should see some console output, and `out.xml` file should get updated (you can check it out by running `cat out.xml`.


Albinutte:
> I didn't do the test steps

image

PR #64, opened 2019-07-22
Add git

### LINKED ISSUE #62: Add git
@RubenAshughyan says:
> Right now I am working on a task feature requested by OnceHub.
> https://github.com/CodeSignal/content/issues/4018
> The special aspect of the task is that the user needs to use git in the solution. Unfortunately, git is not a part of linux as a default, and in order to support git commands while judging tasks in coderunners we need to include git while making docker images.

Seems like we should add it, probably to the base image! That content issue has a hard deadline of July 30-31, so let's try to target July 26 deploy for this.

rowena000:
Closes #62 

### To Test:
I created a coderunner `192.241.161.144` on rundash, you can use it for testing.
1. Go to main app and add `{host: '192.241.161.144'}` to `rundashService::_callRuyndash`
2. Run this script to create a filesystem task:
```js
const taskId = db.tasks.insert({
  "generalType": "algorithmic",
  "sourceName": "bashFilesystemTask",
  "status": "approved",
  "author": "BrainFights",
  "filesystemSupport": true,
  "description": "Finish filling in the server code!",
  "files": [
    {
      "source": "git --version",
      "path": "main.sh"
    }
  ],
  "name": "testingFilesystemTask_codewriting",
  "type": "codewriting",
})
console.log(`go to http://localhost:3000/admin-task/${taskId}, and click run without tests`);
```
3. Running the task above, you should be able to see:
image

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -75,4 +75,5 @@ ADD mongodb.conf /etc/mongodb.conf
 # Install git
 RUN add-apt-repository ppa:git-core/ppa -y \
     && apt-get update \
-    && apt-get install git -y
+    && apt-get install -y --no-install-recommends git \
+    && rm -rf /var/lib/apt/lists/* \
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -73,6 +73,6 @@ ADD mongod.conf /etc/mongod.conf
 ADD mongodb.conf /etc/mongodb.conf
 
 # Install git
-RUN add-apt-repository ppa:git-core/ppa -y
-    && apt-get update
+RUN add-apt-repository ppa:git-core/ppa -y \
+    && apt-get update \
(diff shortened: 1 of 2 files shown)


base/Dockerfile
```diff
@@ -71,3 +71,9 @@ RUN mkdir -p /data/mongodb /data/configdb \
 
 ADD mongod.conf /etc/mongod.conf
 ADD mongodb.conf /etc/mongodb.conf
+
+# Install git
+RUN add-apt-repository ppa:git-core/ppa -y \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends git \
+    && rm -rf /var/lib/apt/lists/* \
```
Albinutte: I think this `\` in the end might break the image.

PR #66, opened 2019-07-24
Acceptance-tests Dockerfile

### LINKED ISSUE #63: Create a Docker image for automated UI tests
https://app.asana.com/0/417895759505117/1132358108136274/f

Preferably not Galen. ~Needs Java, Selenium, some kind of driver -- WebdriverIO is fine, especially since we have experience with it already and it's a JS wrapper (no Java is great, and the sync API is nice too).~

~Consider the kinds of setup we do to get WebdriverIO running on CircleCI -- the need here is similar.~

Maybe need to use Puppeteer? There is an example of what FB is trying to accomplish linked in the Asana task.

Albinutte:
Closes #63.

Creates a new docker image with `puppeteer`, `sinon`, `mocha` and `chai` globally installed.

__How to test this:__

I deployed a new front-end coderunner that you can find on Rundash with IP `192.241.245.30`. SSH into it! Here's what you'll find:
1) If you `cd` into `coderunners` directory, you'll find that it has `Dockerfile` and `test.js`, in addition to the regular files. `Dockerfile` is the file that I added in this PR; `test.js` has the following content:
```js
const puppeteer = require('puppeteer');
const { expect } = require('chai');
const sinon = require('sinon');
const globalVariables = [global.browser, global.expect];

// puppeteer options
const opts = {
  headless: true,
  args: [
    '--no-sandbox',
    '--disable-setuid-sandbox'
  ]
};

describe('sample test', function () {
  let page;

  before(async function () {
    global.expect = expect;
    global.browser = await puppeteer.launch(opts);

    page = await browser.newPage();
    await page.goto('https://www.google.com/');
  });

  it('should have headless chrome of specified version', async function () {
    const browserVersion = await browser.version();
    expect(browserVersion).to.equal('HeadlessChrome/77.0.3844.0');
  });

  it('should have the correct page title', async function () {
    const title = await page.title();
    expect(title).to.eql('Google');
  });

  it('should fail on incorrect page title title', async function () {
    const title = await page.title();
    expect(title).to.eql('Foogle');
  });

  it('should work with sinon', () => {
    const object = { method: function () {} };
    const spy = sinon.spy(object, "method");

    object.method(42);
    object.method(1);

    expect(spy.withArgs(42).calledOnce).to.be.true;
    expect(spy.withArgs(1).calledOnce).to.be.true;
  });

  after(async function () {
    await page.close();

    browser.close();
    global.browser = globalVariables.browser;
    global.expect = globalVariables.expect;
  });
});
```

2) If you run `docker images`, you should see the following:
```
REPOSITORY               TAG                 IMAGE ID            CREATED             SIZE
puppet                   latest              f4bb5424d0e1        9 minutes ago       1.6GB
codesignal/ubuntu-base   v5.0                dc505459ebfd        3 hours ago         1.16GB
codesignal/ubuntu-base   v4.0                dc6b8cfbadc2        6 months ago        1.11GB
codesignal/galen         v1.1                9b799b39bbed        8 months ago        2.03GB
```

`puppet:latest` is the image I built from this new `Dockerfile`.

3) Run `docker run -itv "/root/coderunner":/usercode  puppet:latest /bin/bash` to span a new container using `puppet` image, with `/coderunner` content under the `/usercode` directory in an interactive mode. From there, run something like `mocha usercode/test.js` to get output similar to this one:
```
  sample test
    1) should have headless chrome of specified version
    ✓ should have the correct page title
    2) should fail on incorrect page title title
    ✓ should work with sinon


  2 passing (899ms)
  2 failing

  1) sample test
       should have headless chrome of specified version:

      AssertionError: expected 'HeadlessChrome/77.0.3854.3' to equal 'HeadlessChrome/77.0.3844.0'
      + expected - actual

      -HeadlessChrome/77.0.3854.3
      +HeadlessChrome/77.0.3844.0
      
      at Context.<anonymous> (/usercode/test.js:29:31)
      at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:189:7)

  2) sample test
       should fail on incorrect page title title:

      AssertionError: expected 'Google' to deeply equal 'Foogle'
      + expected - actual

      -Google
      +Foogle
      
      at Context.<anonymous> (/usercode/test.js:39:22)
      at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:189:7)
```

Don't be alarmed by `browser.version` returning `HeadlessChrome...` and not `HeadlessChromium...` or something ! It's [perfectly normal](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#browserversion).

That's it! We can now run tests using puppeteer pointing on both remote and local URLs.


acceptance-tests/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+# Install latest chrome dev package and fonts to support major charsets (Chinese, Japanese, Arabic, Hebrew, Thai and a few others)
+# Note: this installs the necessary libs to make the bundled version of Chromium that Puppeteer installs, work.
+RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \
+    && sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list' \
+    && apt-get update \
+    && apt-get install -y google-chrome-unstable fonts-ipafont-gothic fonts-wqy-zenhei fonts-thai-tlwg fonts-kacst ttf-freefont \
+      --no-install-recommends \
+    && rm -rf /var/lib/apt/lists/*
+
+ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true
```
artahian: I'm not sure I get why this environment variable is set. If puppeteer skips the chromium download, doesn't that mean there will be no chromium to run?

acceptance-tests/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+RUN npm install -g \
+    puppeteer \
+    chai \
+    mocha
```
artahian: Can we add `sinon` too (https://sinonjs.org/)? It was in Facebook's provided example, and it is used to spy on objects/functions to capture the call arguments. It is useful in particular in cases like theirs where we need to mock a tracking function before the page load and then check what arguments it was called with.

acceptance-tests/Dockerfile
Albinutte: We install it separately above; it's just NPM doesn't have the right to install stuff outside of node directory, so things crash hard. Chrome is still there, it's just we need to specify `executablePath: 'google-chrome-unstable',` in puppeteer options (see script above)

acceptance-tests/Dockerfile
Albinutte: Sure! Let me do that.

PR #83, opened 2019-09-04
Create a .NET Core Dockerfile

### LINKED ISSUE #81: Add support for .NET Core SDK
We currently have support for [.NET Framework](https://en.wikipedia.org/wiki/.NET_Framework), but we want to create something that specifically uses [.NET Core](https://en.wikipedia.org/wiki/.NET_Core).

Let's create a Dockerfile for .NET Core that has the dependencies needed to run this task. Contact me for more details on the task.

Make sure xUnit support is included, even though we don't have a unit test adapter for it at the moment.

Also make sure Newtonsoft JSON parsing library is available.

menewman:
Closes #81.

Successful Dockerhub build:
https://cloud.docker.com/u/codesignal/repository/registry-1.docker.io/codesignal/dotnetcore/builds/ac6e8b6c-03b1-4747-b738-dd6c3d396c84

(The last command is `dotnet --version`, which successfully returns `2.2.401`.)

I verified that this is enough to run a simple .NET Core "hello world" example, as well as a slightly more complicated example using packages like xUnit and Json.NET. (If you'd like to test yourself, I can point you to some more specific setup, just don't want to post it here in the PR body since this is a public repo and some of that stuff is a bit internal in nature.)

### DIFF
--- a/dotnetcore/Dockerfile
+++ b/dotnetcore/Dockerfile
@@ -7,13 +7,14 @@ FROM codesignal/ubuntu-base:v5.0
 RUN wget -q https://packages.microsoft.com/config/ubuntu/16.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb
 RUN dpkg -i packages-microsoft-prod.deb
 
-# need this to skip Y/n questions during SDK install
+# skip Y/n questions during SDK install
 ENV DEBIAN_FRONTEND='noninteractive'
 
 # Install .NET SDK
-RUN apt-get install apt-transport-https
-RUN apt-get update
-RUN apt-get -yq install dotnet-sdk-2.2
+RUN apt-get -yq install apt-transport-https --no-install-recommends \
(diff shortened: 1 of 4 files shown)


dotnetcore/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+# need this to skip Y/n questions during SDK install
+ENV DEBIAN_FRONTEND='noninteractive'
+
+# Install .NET SDK
+RUN apt-get install apt-transport-https
+RUN apt-get update
+RUN apt-get -yq install dotnet-sdk-2.2
+
+# Verify
+RUN dotnet --version
```
Albinutte: Can you please run this through https://www.fromlatest.io/#/?

menewman: Thanks -- I didn't even know that was a thing! Also added a link in the README for future Dockerfile editors

PR #85, opened 2019-09-10
Add python libs

### LINKED ISSUE #84: Add libs to python dockerfile
https://app.asana.com/0/417895759505117/1138910775927513/f

Eduard says:
> Currently we are working on one Python file system task for Quora, which requires several libs to be installed on thecoderunners.
> Can we please add the following libs to coderunners ?
> * appdirs
> * attr
> * attrs
> * click
> * toml
> It will be great to have these libs on production by the end of next week (if it's possible).

Dima says:
> To clarify: all these libraries can be installed via pip: 'pip install <lib>' should work for all of them

My question:
> Python 2 or Python 3? Also, why would you need both attr and attrs, instead of just attrs?

Dima's response:
> These libraries are needed for Python 3 only in this particular task, but maybe it's worth installing them for Python 2 as well if they are available for it. Regarding `attr` and `attrs` - I think you're right and just `attrs` should be enough!

ivyc81:
closes #84 

Add
`appdirs`
`attrs`
`click`
`toml`
to both python2 and python3.

### DIFF
--- a/python/Dockerfile
+++ b/python/Dockerfile
@@ -27,6 +27,10 @@ RUN conda create -n py2 -y python=2 \
        MySql-Python \
        SQLAlchemy \
        toolz \
+       appdirs \
+       attrs \
+       click \
+       toml \
     && conda create -n py3 -y python=3 \
        numpy \
        scipy \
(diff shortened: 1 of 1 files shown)


python/Dockerfile
```diff
@@ -39,8 +43,12 @@ RUN conda create -n py2 -y python=2 \
...
+       toolz \
+       appdirs \
+       attrs \
+       click \
+       toml
 
 RUN ["/bin/bash", "-c", "source activate py3 && python -m nltk.downloader popular \
       && rm -rf /var/lib/apt/lists/* /var/log/dpkg.log \
-      && conda clean --all --yes"]
+      && conda clean --all --yes"]
```
Albinutte: Do you mind bringing back newline here?

PR #91, opened 2019-10-08
.NET Core: Install the 2.2 runtime alongside the 3.0 SDK

menewman:
Follow up to the previous PR updating the SDK to 3.0. If we don't explicitly install the 2.2 runtime, it will break existing content that targets 2.2. :(

Successful build here:
https://cloud.docker.com/u/codesignal/repository/registry-1.docker.io/codesignal/dotnetcore/builds/c8731f59-5628-406f-8cfa-5b666435c76c


menewman:
Although my coderunner work hasn't been super productive, I'm still going to request review on this Dockerfile PR since I don't want to leave it in limbo, and I think the image itself is valid.

PR #92, opened 2019-10-31
add futile.logger

### LINKED ISSUE #327: Release second layer of Ubuntu 20 based images part #1
We need to merge `acceptance-tests`, `android-builder`, `cpp`, `cs` dockerfile changes

ivyc81:
added `futile.logger` for better logging
change how we install `tryCatchLog`

Look at the success build on dockerHub https://cloud.docker.com/u/codesignal/repository/registry-1.docker.io/codesignal/r-env/builds/ae579ee8-a4ee-4eb0-ae1a-e25a04adfab0

Also I successfully deployed a coderunner pointing to the new `r-env:dev`. Go to rundash and look for coderunner version `327-update-r`


r/Dockerfile
```diff
@@ -11,7 +11,8 @@ RUN add-apt-repository -y ppa:marutter/rrutter \
     r-cran-dplyr \
     r-cran-magrittr \
     r-cran-data.table \
+    r-cran-futile.logger \
+    r-cran-devtools \
   && rm -rf /var/lib/apt/lists/*
 
-RUN echo "r <- getOption('repos'); r['CRAN'] <- 'https://cloud.r-project.org/'; options(repos = r);" > ~/.Rprofile
-RUN Rscript -e "install.packages('tryCatchLog')"
+RUN Rscript -e "library(remotes); install_github('aryoda/tryCatchLog');"
```
menewman: Can you please explain the rationale behind these changes? In particular...
1. Why include `devtools`?
2. Why change how we install `tryCatchLog`?

ivyc81: It keeps showing
```
Warning message:
package ‘tryCatchLog’ is not available (for R version 3.4.4)
```
When I try to build the docker image wiaht `Rscript -e "install.packages('tryCatchLog')"`

My guess is tryCatchLog is removed from CRAN https://cran.r-project.org/web/packages/tryCatchLog/index.html
So I used the second option 1 from https://github.com/aryoda/tryCatchLog
Screen Shot 2019-11-01 at 1 58 52 PM

menewman: Hm, I see.

I suppose that is acceptable, then, if the reason is that CRAN delisted the package. I just worry that the development version might be too cutting-edge. (When you use the most cutting edge things, sometimes you get cut.)

Left a comment for `aryoda` here, since this seems like something they know about but that kind of got left by the wayside:
https://github.com/aryoda/tryCatchLog/issues/47#issuecomment-548950520

PR #108, opened 2020-02-05
add support for Javafx

### LINKED ISSUE #104: Add support for JavaFX
Blocked by #103. No point in doing this for Java 12 before we've upgraded to support Java 13.

lizhoucheng:
close #104 

## Changes
Now the helper classes from JavaFX are available. But it does not have graphic functionalities.
The changes in coderunner can be found here: https://github.com/CodeSignal/coderunner/pull/434

## Test

1. deploy a coderunner from `javafx` branch (there are existing ones)
2. point the main app to to the coderunner (configured with `RundashAPI`)
3. run the code like this:
```
import javafx.util.*;

class Solution {

    public static void main(String[] args) {
        Pair p1 = new Pair(3,4); 
        System.out.println(p1);
        
        Duration duration = new Duration(0.03);
        System.out.println(duration);
    }
}
```


lizhoucheng:
The code in the main class actually uses the class from `javafx.util`. I add one more example, so that means `Pair` and `Duration` are both the classes from javafx.

PR #119, opened 2020-03-10
Enable lowercase-table-names before installing mysql

Albinutte:
[Here](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_lower_case_table_names) is the explanation of why this solution works.

I will also submit a small PR into the coderunner repo which will contain tests that you will be able to run against the coderunner with IP `67.205.136.94`, which is the one on which I built the `mysql` dockerfile.

Also updated the "show version" test, since mysql version also got bumped a bit.


rowena000:
Sorry I'm confused, `67.205.136.94` doesn't have the updated test does it? How do I run test against it? I deployed a coderunner `192.241.153.214` myself but it didn't pass the `mysql` test.

Albinutte:
@rowena000 nope, it doesn't! It contains only the updated `mysql` image, as well as updated `compilers.js`. Please check out https://github.com/CodeSignal/coderunner/pull/463#issuecomment-597761279 for detailed instruction.

Albinutte:
Yey!

PR #125, opened 2020-03-24
Auto build dev on docker

ivyc81:
### Functional
Now circleci will build updated dockerfiles and push it to dockerhub with dev tag. So if you need to update many images at once, you don't need to go to each repo and configure custom build rules.
Check should fail if an image fails to build. Look at [this commit](https://circleci.com/gh/CodeSignal/dockerfiles/16). I broke the acceptance-tests dockerfile on purpose.
Check should fail if pushing to dockerhub fails. Look at [this run](https://circleci.com/gh/CodeSignal/dockerfiles/17) where I don't have permission to push to elixir repo.

New workflow if you need to updated image
1. Update docker file(s)
2. Create a PR, cicrleci will automatically build and push the image(s) to dockerhub with `dev` tag
3. You can go to circleci to look at build log to check whether build succeeded
4. Deploy a develop coderunner pointing to the image(s) with `dev` tag
5. If an image has dependencies, after releasing a new version of the image, repeat the process again for its dependencies

## Testing
I made some changes in elixir, r, d and acceptance-test.
1. Go to Circleci
2. Look at the tests. There should be build logs for elixir, d, r and acceptance-tests.
3. Rerun  workflow.
4. Go to dockerhub. Should see codesignal/elixir, codesignal/r-env, codesignal/d-env and codesignal/acceptance-tests dev tag updated recently.

* I will revert the dummy changes I made before merging. 
* Right now the credential saved as environment variables `DOCKERHUB_USERNAME` and `DOCKERHUB_PASSWORD` is my login. Maybe we should create a codesignal account for that purpose.


.circleci/config.yml
```diff
@@ -0,0 +1,32 @@
...
+            echo "$DOCKERHUB_PASSWORD" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
+            echo List of updated files
+            FILES=$(git diff-tree --name-only -r origin/master HEAD)
+            for F in $FILES; do echo $F; done
+            REGEX="(.*)\/Dockerfile"
+            for F in $FILES
+            do
+              if [[ $F =~ $REGEX ]]
+              then
+                IMAGE=${BASH_REMATCH[1]}
```
Albinutte: Sorry for PR stalking, but please note that this won't work for a few languages that have dockerfiles stores in directories that don't match repo names on dockerhub. Those are the languages that have one-letter names, like `D` and `R`. The reason for mismatch is that dockehub has a minimum limit on the number of characters in the repo title, so for those languages the repos are called `d-env` and `r-env`. There might be other mismatches that I don't remember off the top of my head.

ivyc81:
@menewman I updated the PR description to better document how this PR is suppose to solve the issue. Let me know if there's more clarification needed.

.circleci/config.yml
ivyc81: Ah I see. I will do something about it.

ivyc81: @Albinutte I'm a little confused with julia & julia-1. We build `julia` dockerfile to `julia-1` repo on dockerhub, right?

List of mismatch I found so far
|Dockerfile|dockerhub|
|----------|-----------|
|d               |d-env         |
|r                |r-env          |
|julia          |julia-1          |

Albinutte: julia-1 is deprecated, and we should be able to delete it once the latest version of the coderunners are deployed. D and R seem to be the only languages with the mismatch then 👍

ivyc81:
> Is there any particular configuration needed to set this up?

I'm thinking create a new dockerhub account that also has developer access. And use the login here to replace mine.

Screen Shot 2020-03-31 at 10 57 04 PM

menewman:
@botty-mcbotface is now the proud owner of the Dockerfiles credentials in CircleCI

Albinutte:
I'm going to hit dat big green button and let the CI handle my OpenCV image 🤞

Albinutte:
Oh shoot, I missed the UNDO CHANGES changes 🤦‍♂ Sorry!

PR #127, opened 2020-03-25
Added postgresql and mysql support for R and python

eduard-piliposyan:
There is no issue at this moment for this. But for some client task we need to support postgresql and MySQL support. This change adds missing libs for working with MySQL and PostgreSQL in R and Python.


eduard-piliposyan:
@menewman actually I  build docker image locally but can build in dockerhub as well.

eduard-piliposyan:
@menewman I decided to remove one of the libs for python. Actually, it was the same psql support, but a different driver. So after the build failed I decided to remove that one. 
Now both builds are successful in dockercloud.

PR #129, opened 2020-04-02
OpenCV

Albinutte:
This PR adds a brand new OpenCV image. You can find the image on [dockerhub](https://hub.docker.com/layers/codesignal/opencv/v5.5/images/sha256-69f3ad013d820aea51a38a44fe76aa51391158d93b7af31417ed735a917f7c04?context=repo), pushed there manually. 

Please ignore CI reports. The C++ was built because of my initial version was just its extension (I've deleted it already), and automated OpenCV build probably failed because dockerhub couldn't handle it. I built the image manually and pushed it to dockerhub, and it's working, I guarantee it.

You can find proof in the corresponding PR in the coderunner repo: https://github.com/CodeSignal/coderunner/pull/486. There is a coderunner deployed from the production coderunner as well.


Albinutte:
Ah, looks like the build failed because I didn't give developers `admin` access to the newly created `opencv` repo. It should be fixed now, but in the meantime I triggered the build through the automated builds.

PR #146, opened 2020-05-08
Ruby version update: 2.7.1

menewman:
Updated Ruby version from 2.5 to the latest version (2.7.1) using a more standard installation method: [`ruby-build`](https://github.com/rbenv/ruby-build). Previously we were using a package maintained by the company [Brightbox](https://www.brightbox.com/docs/ruby/ubuntu/), for reasons that are not entirely clear to me. Not that there's anything wrong with Brightbox, but I feel better using something that feels more standard and less proprietary.

Often `ruby-build` is used in conjunction with `rbenv`, which is similar to `nvm` in that it's a way to manage installation of multiple Ruby versions on the same machine. Since we don't really need that capability and installing `rbenv` was a bit of a pain, I eventually opted for the standalone version of `ruby-build`, which works just fine as long as we take care to update the `PATH` appropriately.

Proof of successful build: See the built Docker image on CircleCI! The version is indeed 2.7.1, and gems were also installed successfully:
Screen Shot 2020-05-07 at 7 25 55 PM

You can also:
- Run the Dockerfile through [this opinionated Dockerfile linter](https://www.fromlatest.io/#/) to see if it has any feedback. (It doesn't!)
- Look at my related coderunner PR, which includes a little more detail about how to test this on a coderunner and/or against our main application.


ruby/Dockerfile
```diff
@@ -1,8 +1,24 @@
...
+RUN apt-get -qq update \
+    && apt-get install --no-install-recommends -y autoconf bison build-essential libssl-dev libyaml-dev libreadline6-dev zlib1g-dev libncurses5-dev libffi-dev libgdbm3 libgdbm-dev libdb-dev libpq-dev \
     && rm -rf /var/lib/apt/lists/*
+
+# Install ruby-build
+RUN git clone https://github.com/rbenv/ruby-build.git
+RUN PREFIX=/usr/local ./ruby-build/install.sh
+
+# Build Ruby
+RUN ruby-build 2.7.1 /usr/local/ruby-2.7.1
```
Albinutte, line 14: @menewman is there a reason to have multiple layers in this docker files? For commands this small as the ones above, running them in a single layer should help with the build time of the image. I would strongly suggest squashing the four `RUN` command above into one.

ruby/Dockerfile
```diff
@@ -1,8 +1,24 @@
...
 
-RUN apt-add-repository ppa:brightbox/ruby-ng \
-    && apt-get -qq update \
-    && apt-get install --no-install-recommends -y ruby2.5 \
-    && apt-get install --no-install-recommends -y ruby`ruby -e 'puts RUBY_VERSION[/\d+\.\d+/]'`-dev libpq-dev \
-    && gem install rspec pg mysql2 activemodel activerecord activesupport json faker pry \
+# Install required dependencies for Ruby and ruby-build
+# https://github.com/rbenv/ruby-build/wiki#ubuntudebianmint
+RUN apt-get -qq update \
+    && apt-get install --no-install-recommends -y autoconf bison build-essential libssl-dev libyaml-dev libreadline6-dev zlib1g-dev libncurses5-dev libffi-dev libgdbm3 libgdbm-dev libdb-dev libpq-dev \
```
Albinutte, line 6: Not a huge deal, but this command can be made a bit more readable with every package on a new line, as, for example, in the Java Dockerfile:
https://github.com/CodeSignal/dockerfiles/blob/30ba37fbadf1de411cc8191fc6d8bf10a42c1525/java/Dockerfile#L3-L13

ruby/Dockerfile
menewman: Nope, I was just being careless and writing this Dockerfile as if it were a Bash script (since I was referencing Bash install instructions as I wrote it). You're absolutely right; I should combine these instructions and minimize the number of layers. Thanks! 👍

ruby/Dockerfile
menewman: That's true; will update!

menewman:
Thanks for the review and the gentle "this works" on my slightly ugly Dockerfile! 😅 Hopefully it's a little better now. I don't think the build time is much changed (still 6-7 minutes either way), but it's more readable now.

PR #147, opened 2020-05-11
Adding new libs for python

ArDrambyan:
Adding new python libs for https://github.com/CodeSignal/coderunner/issues/540

## Now added 
- `spacy`
- `gensim` 
- `matplotlib` 
- `seaborn` 
- `pytest` 
- `pymongo` 
- `flask` 
- `flask-sqlalchemy` 
- `flask-migrate`
- `requests` 
- `ipywidgets` 
- `pandas-profiling` 
- `plotly`
- `tweepy`
- `eli5`
- `category_encoders `
- `pdpbox`
- `shap`
- `xgboost`
- `en_core_web_md` model for `Spacy`
- `tensorflow` 2.1 version
- `basilica`
 

Coderunner's tests are passing with these changes, you can check the coderunner with `Aram's experimental` label in rundash


ArDrambyan:
Hi @menewman could you please review my changes? I couldn't find `basilica` package in [anaconda.org](https://anaconda.org/search?q=basilica). Maybe we should install it using pip?
Because I'm new in this repo I not sure do we need to split this image into 2 different images.

ArDrambyan:
I agree with you but I don't know how to split these libs 😄 I guess that changes will lead to changes in the `codesignal` and in the `coderunner`?

eduard-piliposyan:
> Looks like adding some of these libs (most likely the data science-related ones) nearly doubled the image size, which is pretty wild.
I definitely would like to keep the Python image size down. Maybe it's time to consider exploring a dependent image that contains data science models and derives from Python so that we don't have to make Python itself enormous?

@menewman I completely agree that having a separate docker image for data science makes everything more structured. I agree that upload/download of a docker image to/from dockerhub will be slower for bigger images. But does it also have a physical(e.g performance) meaning? Does a docker image with a bigger size always use more RAM? Or is the startup slower? I didn't do any benchmarking on this, but AFAIR current docker performance isn't faster from the times that we had just one dockerfile. Of course, coderunner deployment is way faster than it was before and that's a huge benefit. 

But the issue that I see with many hierarchical dockerfiles is the **maintenance**. Changing even the list level dockerfile requires changes in 2 repos, build in dockerhub, upload of the new image to dockerhub, coderunner release creation and deploy. In case of changes in higher levels of the dockerfile tree, we will have more steps and this results in a very painful process (especially for the base image).
That why I would like to keep tree levels as small as possible, and even spend some time to come up with a better way of organizing dockerfiles which will speed up the maintenance process of coderunners. Of course, this is a different discussion thread.

eduard-piliposyan:
And if the size of the image is important and we need to split it we can also consider splitting it to py2 and py3 images. That won't require any code change on the codesignal side and won't depend on task type.

menewman:
## To Aram
> I agree with you but I don't know how to split these libs 😄 I guess that changes will lead to changes in the `codesignal` and in the `coderunner`?

Yes, although I don't think the changes would be enormous. The change in `codesignal` would be to allow targeting a "language" or environment for run requests that isn't just `py3` but is `py3-data-science` or something like that (exact semantics not set in stone). The change in `coderunner` would be to accept that value and point to the right Docker image.

> Maybe it makes sense to keep all in a single image for this task open another issue for splitting to have support ASAP?

Perhaps, although I worry that this could turn into one of those "oh, we'll refactor it later" things that never gets done. You know how it goes! 😅 

If we already know the installation steps, then splitting the Dockerfiles won't take that much longer. It just requires creating a new "repo" in DockerHub and then updating the coderunner to point there.

## To Eduard
@eduard-piliposyan -
> Does a docker image with a bigger size always use more RAM? Or is the startup slower?

Yes, [Docker's docs](https://docs.docker.com/develop/dev-best-practices/#how-to-keep-your-images-small) do seem to say that startup is slower for larger images:
> Small images are [...] faster to load into memory when starting containers or services.

So it seems like going from ~2 to ~4 GB image size would have an impact on container startup time, and therefore on our app's code total run time.

It might also affect the memory load of keeping Python images in our Docker pool? I haven't looked at the pool code closely, but maybe @ArDrambyan could comment on whether Docker image size makes a big difference there or whether we can benchmark before/after the changes in this PR. (Also CCing @Albinutte, who reviewed that PR)

> But the issue that I see with many hierarchical dockerfiles is the **maintenance**. Changing even the list level dockerfile requires changes in 2 repos, build in dockerhub, upload of the new image to dockerhub, coderunner release creation and deploy. In case of changes in higher levels of the dockerfile tree, we will have more steps and this results in a very painful process (especially for the base image).

A few things:
- Whether we have a separate data science image won't impact how many repos we need to make changes in.
- Whether we have a separate data science image won't require any additional work to be done in DockerHub (we simply need to push tags in this repo and DockerHub builds automatically).
- We should not have to manually upload any new images to DockerHub.
- Whether we have a separate data science image won't impact our coderunner release creation/deployment steps in any way.
- The release process is already multi-step, so this doesn't really seem to add a new "level" compared to what we already have with, say, the Java image and its dependencies.
- The development process for updating dependent Docker images, although not perfect, is much easier now after #125.

> That why I would like to keep tree levels as small as possible, and even spend some time to come up with a better way of organizing dockerfiles which will speed up the maintenance process of coderunners. Of course, this is a different discussion thread.

I don't really think it is a different discussion thread since organization of Dockerfiles is very pertinent to this PR!

> And if the size of the image is important and we need to split it we can also consider splitting it to py2 and py3 images. That won't require any code change on the codesignal side and won't depend on task type.

I'm also generally supportive of that change but it doesn't really solve the problem of large data science dependencies making "regular Python coding" (without data science) slower. Also depends on whether splitting the two Python languages into different images would result in image size being halved or not (hopefully it would but I think it depends on whether they currently share any dependencies across their data science libs).

ArDrambyan:
Thank you, Michael, for explanations, as a temporary solution I'm going to keep all libs in `codesignal/python` image and I think it won't hit performance, because of the docker pool. Then I'll move data science libs into another image and after that, we will allow the `docker pool` only for B2B so we will avoid any kind of performance hit.

eduard-piliposyan:
@menewman thanks for the detailed answer.

> The release process is already multi-step, so this doesn't really seem to add a new "level" compared to what we already have with, say, the Java image and its dependencies.
In the case of inheriting `py-data-science` from `py` for every change in python dockerfile, we will require to make one additional step of changes (PR) for inherited dockerfiles. Yes similar to Java and it's dependencies. But at least for python, it was simpler, not it will also become multistep.

> It might also affect the memory load of keeping Python images in our Docker pool? I 
It will affect because we would like to keep separate containers for python and separate containers for python-data-science. So it will double the used memory.

> I'm also generally supportive of that change but it doesn't really solve the problem of large data science dependencies making "regular Python coding" (without data science) slower. Also depends on whether splitting the two Python languages into different images would result in image size being halved or not (hopefully it would but I think it depends on whether they currently share any dependencies across their data science libs).
Because py2 and py3 have different installation flow for all libs I believe they don't share much and it will be halved. 

One more note about Data Science tasks I forgot to mention in my previous comment. We almost finished the integration of Jupyter notebook and as a result, most of the data science users will switch to it (I believe most of them use it during the interviews). And also I believe that after persistent sessions we will setup our own jupyter server, and at that moment we can prepare a separate data science dockerfile, which will also include jupyter inside.

menewman:
Thanks for the thoughtful discussion, both!

## To Aram
> Thank you, Michael, for explanations, as a temporary solution I'm going to keep all libs in `codesignal/python` image and I think it won't hit performance, because of the `docker pool`. Then I'll move data science libs into another image and after that, we will allow the `docker pool` only for B2B so we will avoid any kind of performance hit.

Yep, for this PR, let's move forward with the mega-image for the short-term fix and consider splitting as a follow-up, since we do have a short timeline in mind for adding these libs. 👍 

## To Eduard
> In the case of inheriting `py-data-science` from `py` for every change in python dockerfile, we will require to make one additional step of changes (PR) for inherited dockerfiles. Yes similar to Java and it's dependencies. But at least for python, it was simpler, now it will also become multistep.

Yes, it's true that creating a new data science image derived from Python adds another hierarchical layer, but the change here will be small and I think it's worth it if we can keep the size and startup time of normal Docker containers optimized. The update process for the dependency is actually pretty simple as well (just changing a single line), so even though it's another step, it shouldn't be an overly onerous one. But of course it's always going to be a tradeoff.

> It will affect because we would like to keep separate containers for python and separate containers for python-data-science. So it will double the used memory.

Hm, interesting way of looking at it! That's assuming we'd want the same number of Data Science containers in the pool as regular Python containers, in which case you're right that it would increase the size (likely not by double since the data science libs have most of the size, but I get what you mean). That said, I think we wouldn't need to maintain the same size pool for the Data Science image since 1) soon we'll have Jupyter anyway, and 2) data science tasks should be used less frequently than "regular" Python. So ultimately I do think it would lead to a improvement in pool memory reduction if we could split them.

> Because py2 and py3 have different installation flow for all libs I believe they don't share much and it will be halved. 

Hopefully that's true, and I do agree it would be nice to split the two languages. Especially since Python 2 won't be receiving updates anymore; it would be great to be able to update Python 3 in isolation. That said, it's not a mutually exclusive change with splitting out data science libs.

> One more note about Data Science tasks I forgot to mention in my previous comment. We almost finished the integration of Jupyter notebook and as a result, most of the data science users will switch to it (I believe most of them use it during the interviews). And also I believe that after persistent sessions we will setup our own jupyter server, and at that moment we can prepare a separate data science dockerfile, which will also include jupyter inside.

I agree that adding Jupyter will help, for sure, since hopefully it will allow us to push back more on requests to add big data science libs to our images.

python/Dockerfile
```diff
@@ -9,16 +9,17 @@ RUN apt-get -qq update \
     && conda update conda \
     && apt-get -qq -y remove bzip2 \
     && apt-get -qq -y autoremove \
-    && apt-get autoclean \
+    && apt-get autoclean \ 
+    && apt-get -qq -y install python-pip \
```
menewman, line 13: Albina showed us a pretty cool tool for linting Dockerfiles:
https://www.fromlatest.io/#/

It's nice to run Dockerfiles through this to see if it has any suggestions! Here, there's one small suggestion:
> Consider using a --no-install-recommends when apt-get installing packages. This will result in a smaller image size.

PR #152, opened 2020-05-19
updated python's docker image version to 3.8

### LINKED ISSUE #151: Update conda's python3 environment to 3.8.2
We also need to update `conda`'s environment to be able to use python3.8

ArDrambyan:
Closes #151 

## Changes 
Updated conda's python3 version to 3.8.2.


ArDrambyan:
@menewman I think I should create 5.8.1 version, right?

ArDrambyan:
It seems like part of our packages doesn't support python 3.8, that's why `conda` uses 3.7

menewman:
Yeah, we'll need to publish a new Python release, like 5.8.1, once these changes are ready.

> It seems like part of our packages doesn't support python 3.8, that's why `conda` uses 3.7

Hm... so you mean that we just shouldn't update to 3.8 at this time?

ArDrambyan:
> Hm... so you mean that we just shouldn't update to 3.8 at this time?

I think so, I couldn't even update python version after installing packages with 3.7

menewman:
Hm, that's not a super satisfying outcome since Python 3.8 has been out since last October (when will these packages update?) but it's probably okay to leave it at 3.7 for now given that there's not a specific urgent need for the latest version.

We'll need to be able to bump it up at some point, though. 🤔

ArDrambyan:
I think we can try to update python's version when splitting, create one conda's python environment for simple algorithmic tasks with 3.8 and one for data science with 3.7

menewman:
That sounds like a reasonable plan to me for now!

PR #157, opened 2020-06-02
Splitting python image

### LINKED ISSUE #155: Split python image
@ArDrambyan commented on [Tue Jun 02 2020](https://github.com/CodeSignal/codesignal/issues/17062)

We need to split the `codesignal/python` image into `codesignal/python2` and `codesignal/python3` images and also replace `conda` installations with `pip` to decrease execution time.

ArDrambyan:
Closes #155 

## Changes
 - splitting python image into two images: python2, python3
- all packages are installed using pip instead of conda


ArDrambyan:
Hi, @menewman @Albinutte can you please review my changes? Also, CI tests are throwing error, because I removed python directory,

ArDrambyan:
For testing, you can use coderunner: 142.93.50.223, which is marked as `Aram's`. We are now supporting all libs that we asked for (we don't support only `category_encoders` for py2, because this lib doesn't have any version that supports py2). I'm using this code for initial testing:
```
import pathlib 
import numpy 
import scipy 
import pandas 
import sklearn  
import nltk 
import requests 
import configparser 
import mysql.connector
import mysql 
import sqlalchemy 
import psycopg2 
import toolz 
import appdirs 
import attr
import click 
import toml 
import spacy 
import matplotlib 
import seaborn 
import pytest 
import pymongo 
import flask 
import flask_migrate 
import ipywidgets 
import pandas_profiling
import plotly 
import tweepy 
import eli5 
import pdpbox 
import shap 
import xgboost
import basilica
import tensorflow
import tensorflow.keras
import gensim
import category_encoders

nlp = spacy.load("en_core_web_sm")

print('Hello!')
``` 
(don't forget to remove `category_encoders` and change print command for testing py2)
Execution of this code may take a bit longer with this libs, but code with our default imports runs faster than ever.

python3/Dockerfile
```diff
@@ -0,0 +1,69 @@
...
+      pdpbox \
+      shap \
+      xgboost \
+      basilica \
+      tensorflow \
+      gensim \
+      category_encoders && \
+      python3 -m spacy download en_core_web_sm
+
+RUN rm -rf temp
```
Albinutte, line 69: Please try to keep creating/deleting artifacts as close as possible, preferably in a single layer (`RUN` command).

In this case, it would be nice to do something like this:
```dockerfile
RUN mkdir temp \
  && cd temp \
  && ...install python \
  && ...configure python \
  && ...install pip \
  && ...cleanup
```

Ideally, you want every layer to be as independent as possible, while also keeping its size to the minimum.

ArDrambyan: > Yes! Please update `python2` and `python3` to get at least two automated builds configured, so that dockerhub could build new images based on tags in this repo. You can see examples [here](https://hub.docker.com/repository/docker/codesignal/python/builds), and format description in README.md.

I think DockerHub is fully configured, cause automated builds are passing and we have python2:dev and python3:dev in our DockerHub

ArDrambyan:
> Yes! Please update `python2` and `python3` to get at least two automated builds configured, so that dockerhub could build new images based on tags in this repo. You can see examples [here](https://hub.docker.com/repository/docker/codesignal/python/builds), and format description in README.md.

I think DockerHub is fully configured, cause automated builds are passing and we have python2:dev and python3:dev images in our DockerHub

ArDrambyan:
I was wrong, it wasn't. But I think it should be okay now

.circleci/config.yml
```diff
@@ -17,7 +17,7 @@ jobs:
             echo "$DOCKERHUB_PASSWORD" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
 
             echo List of updated files
-            FILES=$(git diff-tree --name-only -r origin/master HEAD)
+            FILES=$(git diff-tree --name-only -r --diff-filter=ACMRT origin/master HEAD)
```
menewman, line 20: 🎉

PR #159, opened 2020-06-08
P.Session: Support nvm and node 10

### LINKED ISSUE #158: Support node 10 in Ubuntu base
Is required for this big issue
https://github.com/CodeSignal/coderunner/issues/570

salbert96:
Closes #158 
I checked and session-server works with this ubuntu base.
Note that we install node 10 but the system default will remain the one that was already installed so the changes are backward compatible. 
Once we create a new image for ubuntu base I will create more PRs and releases for the other dependencies.


salbert96:
Apparently there was no need for the script 🤦  we could just do global replace. But the script might be useful in the future when we want to update the third layer in one batch.

_bin/update_dependencies.py
```diff
@@ -0,0 +1,111 @@
+import sys
```
menewman, line 1: This is somewhat my personal preference, but since the majority of our developers use JS as their primary language and Python not nearly as frequently (or not at all), I would rather have scripts written in Node since such scripts will be easier to read and maintain in the future

base/Dockerfile
```diff
@@ -13,6 +13,9 @@ RUN apt-get clean \
       build-essential \
       curl \
       wget \
+      xz-utils \
```
menewman, line 16: Why are these 3 libs needed?

base/Dockerfile
```diff
@@ -86,4 +89,14 @@ RUN echo "deb http://apt.postgresql.org/pub/repos/apt/ xenial-pgdg main" >> /etc
        postgresql-contrib-12 \
        postgis \
        postgresql-12-postgis-2.5 \
-    && rm -rf /var/lib/apt/lists/*
\ No newline at end of file
+    && rm -rf /var/lib/apt/lists/*
+
+# Installing nvm
+RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.3/install.sh | bash
```
menewman, line 95: It seems a little odd to install NVM after explicitly installing Node 12 above. If we are going to use NVM anyway we should install it first and use it to install both Node 12 and Node 10 if needed, right?

base/Dockerfile
salbert96: Those are needed for successfully building session-server.

salbert96:
> could you help me understand why we need to support a _lower_ version of Node here than our usual one?

Node 10 is the only version that can be used to build session-server successfully.

base/Dockerfile
salbert96: If we install node 12 with nvm we would have to write `. /root/.nvm/nvm.sh && nvm use 12` with every time we run a command on the docker. But with this approach, everything stays the same but when you want to use node 10 for the session-server you run `. /root/.nvm/nvm.sh && nvm use 10.15.3` inside the docker command. I know it's strange in general but this is because of docker. Maybe there is a workaround but I don't know any yet.

_bin/update_dependencies.py
salbert96: Please don't make me write this on JS 😅  usually when it comes to writing such scripts python the first choice of language. I don't think this script is going to be used by anyone but me for the near future 😃  but having it in python seems very clear. It's like replacing bash scripts with JS.

base/Dockerfile
salbert96: Having node 12 installed in a natural way makes sure that everything will work as before, while I can't guarantee it with nvm.

_bin/update_dependencies.py
salbert96: Or at least I would prefer to have it this way for now and change it to JS later to not spend the same time on it by writing it on JS. We can do it after persistent sessions.

_bin/update_dependencies.sh
```diff
@@ -0,0 +1,47 @@
...
+        ;;
+        # Swallow all unknown parameters
+        *)
+        shift
+        shift
+        ;;
+    esac
+done
+
+python3 update_dependencies.py ${BASE_IMAGE} ${NEW_VERSION}
```
Albinutte, line 47: Writing a bash script to work with a python script seems unnecessary. Instead, you can make python script executable by adding a `#!/usr/bin/env python3` shebang, and running `chmod +x update_dependencies.py`.

You can then use [argparse](https://docs.python.org/3/howto/argparse.html) to provide a nice interface if you feel like it.

_bin/update_dependencies.py
Albinutte: TBF though, this update can be done with a single regular expression in any IDE, so if it's your personal script, perhaps there's no need to commit it? You can add it to your personal gitignore under `$GIT_DIR/info/exclude`.

Otherwise it should be written in a way that could be maintainable by other developers in our team.

salbert96:
Thanks, @Albinutte for the python review 🙂  I will add it to my local gitignore though, and we can implement JS version later if needed.

base/Dockerfile
salbert96: Here is a small video that demonstrates my reasoning 
https://drive.google.com/open?id=1r3sMZtZkekgJO_7MUKTUc5bl7GKP1QzA
When you run this command 
`docker run -it codesignal/ubuntu-base:dev`
Everything will work fine and node version will be okay but when I run this
`docker run codesignal/ubuntu-base:dev /bin/bash -c "node --version" `
It prints 
`/bin/bash: node: command not found`
While if I run this
`docker run codesignal/ubuntu-base:dev /bin/bash -c ". ~/.nvm/nvm.sh && node --version" `
It prints 
`v12.9.0`
So that's why I don't want to use nvm for the first case and use nvm for the second.

salbert96: I think the reason is explained here
https://stackoverflow.com/a/38554839

salbert96: So my suggestion is to use `. /root/.nvm/nvm.sh && nvm use 10.15.3` when a command is related to session-server instead of changing this behavior for running test cases for JS and maybe front-end tasks.

salbert96:
@menewman @Albinutte I removed the shell script but did not change the part related to nvm. The reasoning is explained in the comments.
In general, I am okay to install node 12 with nvm but we will need to add this command `. /root/.nvm/nvm.sh` to all the container commands that use node. Which is pretty risky.

base/Dockerfile
menewman: I agree with Albina's comment; this seems like a solvable setup issue. In the StackOverflow answer you linked, there are several variations of answers that involve figuring out how to source NVM correctly so that `nvm` and `node` can be run correctly.

If possible I think it really would be preferable to use NVM for all Node versions we have installed, since that is the entire purpose of NVM -- it feels unnatural to me to have a standalone Node installation alongside an incompletely configured NVM installation instead of just setting up NVM all the way. Plus, getting NVM properly set up may allow us to more easily do things like have the Node version change in different dependent images (especially the JS image) without having to update the base image.

_bin/update_dependencies.py
menewman: @salbert96 - This is now no longer a current discussion thread since you removed the scripts, and I never would have made you re-write your script in Node, but I would like to point out that there's a connection between these 2 statements:

1.
> [...] since the majority of our developers use JS as their primary language and Python not nearly as frequently (or not at all), [Python scripts are more difficult] to read and maintain in the future

2.
>  I don't think this script is going to be used by anyone but me for the near future

The first fact makes the second statement much more likely to be true!

Also...
> Or at least I would prefer to have it this way for now and change it to JS later to not spend the same time on it by writing it on JS. We can do it after persistent sessions.

It's hard for me to ever be convinced by this argument since usually "do it later" means "do it never" 😅 

But anyway, let's resolve this thread 👍

base/Dockerfile
```diff
@@ -13,18 +13,30 @@ RUN apt-get clean \
...
+
+ENV NVM_DIR /root/.nvm
+
+# Installing Node 12.16.1 with nvm
+
+ENV NODE_VERSION 12.16.1
+
+RUN . $NVM_DIR/nvm.sh && nvm install $NODE_VERSION && nvm alias default $NODE_VERSION && nvm use default
+
+ENV PATH $NVM_DIR/versions/node/v$NODE_VERSION/bin:$PATH
```
salbert96, line 39: This line did the trick

salbert96:
It seems like adding the path did the trick. Although nvm is not being sourced we still have node version working
Screen Shot 2020-06-10 at 12 09 47
If we want to use nvm we still need to write 
`. $NVM_DIR/nvm.sh`
When we run in non-interactive mode.

base/Dockerfile
```diff
@@ -13,18 +13,30 @@ RUN apt-get clean \
       build-essential \
       curl \
       wget \
+      apt-transport-https \
+      apt-utils \
```
salbert96, line 17: These two lines were needed since I was getting this weird error
Screen Shot 2020-06-10 at 12 00 28
The solution was written here
https://unix.stackexchange.com/questions/263801/apt-get-fails-the-method-driver-usr-lib-apt-methods-https-could-not-be-found

base/Dockerfile
salbert96, line 39: If we would like to use a different node version in a different docker image we can copy these 3 lines

salbert96:
This will have ubuntu-base version `v6.0` since it is not backward compatible.
Previously we needed to run `nodejs` command instead of `node`

PR #170, opened 2020-06-11
Changed node path

salbert96:
This was causing error for js test so globally installed packages were not working.

Honestly, I think this should be in ubuntu-base, but it was pre-existing so changed it here.

In case you want me to change it in ubuntu-base we can do it too (although I already created all the release) or we can have it here and change it later when we have all working in production.

I guess for me would be easier to just change JS image, but I am open to suggestions.


menewman:
> In case you want me to change it in ubuntu-base we can do it too (although I already created all the release) or we can have it here and change it later when we have all working in production.

Since this isn't your priority right now and you want to release the images we've already built, changing in the JS image is okay for now. 👍  Especially since we already define this here, even before your latest changes.

But if you think it makes sense to put this in the base image, then could you perhaps create an issue tracking that so that we have a TODO for later? (Or a PR if you're up for it.)

salbert96:
Thanks 👍

PR #173, opened 2020-06-22
Add JS libs: axios, node-fetch, rambda, futil

### LINKED ISSUE #172: Add node-fetch, axios to the JS Dockerfile
Now that the JS `request` library [has been deprecated](https://github.com/request/request/issues/3142) (as of Feb 11 this year), we should make sure to include maintained alternatives.

The two strongest contenders seem to be `node-fetch` (meant to simulate the browser `fetch` API implemented in Node) and `axios` (another very popular request-making library):
https://github.com/node-fetch/node-fetch
https://github.com/axios/axios

menewman:
Closes #172.

Simply adds node-fetch and axios to the list of globally available Node modules. The proof this works should be in the successful build.

https://www.npmjs.com/package/node-fetch
https://www.npmjs.com/package/axios


menewman:
Relevant excerpt from successful build:
Screen Shot 2020-06-22 at 12 16 44 PM

menewman:
We also received a request to add `rambda` and `futil` (which are functional programming libraries), and that seemed reasonable to do at the same time, so I added those too.

PR #179, opened 2020-07-05
Fixing Tensorflow installation

### LINKED ISSUE #176: Tensorflow error in Python2 image
Here is what I see in dockerhub for the latest built python 2 image

I think this was introduced when we separated python2 and python3

The best solution would be to fix the versions of pip packages inside dockerfiles.

ArDrambyan:
Closes #176 

## Changes
- Installed extra lib for `tensorflow` (for python2/python3)
- specified versions for some packages to make `tensorflow`'s installation successful 
- downgraded `tensorflow` to 2.0 because I faced this issue: https://github.com/tensorflow/tensorflow/issues/35968


ArDrambyan:
Hi @menewman, I tried to install `libnvinfer6` and another solution from that thread but I wasn't able to install that libs.

menewman:
@ArDrambyan - I think that's because you have to add the NVIDIA package repositories first before trying to install the package, otherwise it won't be found:
https://www.tensorflow.org/install/gpu#ubuntu_1604_cuda_101

ArDrambyan:
@menewman - Yes I used that example and I'm getting this error: "error processing archive /var/cache/apt/archives/libcudnn7_7.6.4.38-1+cuda10.1_amd64.deb (--unpack):
 cannot copy extracted data for './usr/lib/x86_64-linux-gnu/libcudnn.so.7.6.4' to '/usr/lib/x86_64-linux-gnu/libcudnn.so.7.6.4.dpkg-new': failed to write (No space left on device)"

PR #187, opened 2020-07-28
Bumped languages based on ubuntu-base: v7.0

Albinutte:
A follow-up to #186, bumps images based on `ubuntu-base`.


Albinutte:
> Why v7.0? Is PowerShell a breaking change?

No, but I find it easier to version images by the dependency depth. I.e., updating a base image always bumps major version, updating C++/Java/... always bumps minor version, and updating Scala/... always updates patch version. I'm too lazy to go through all images and search for the next one that will work for everything otherwise 😅

menewman:
Haha, okay, fair enough!

PR #189, opened 2020-08-10
add vim and gdb

### LINKED ISSUE #183: P.Session: Install vim and GNU Debugger
We can install vim and GNU Debugger on `ubuntu-base` so that they can be used from the terminal without requiring additional installations. @menewman this is not a high priority one but I would love to know your opinion.

lizhoucheng:
close #183 

## Change
Add vim and gdb to base image

## Test
The image is successfully built. I also update JS image to use the new base image so you can test. The coderunner deployed is `167.99.10.69`.
To test, 
1. set the `targetHost` in rundashApi in the main app as `167.99.10.69`
2. open the task `terminalDemo` and switch to Javascript language
3. you can type `vim` to use vim or `gdb` to use the gnu debugger in the terminal.


js/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM codesignal/ubuntu-base:v7.0
+FROM codesignal/ubuntu-base:vim-gdb
```
salbert96, line 1: Tagging this since it should be deleted.

js/Dockerfile
```diff
@@ -29,5 +29,3 @@ RUN npm install -g --unsafe-perm=true \
     validator  \
     when \
     ws
-
-ENV NODE_PATH=$NVM_DIR/versions/node/v$NODE_VERSION/lib/node_modules
```
salbert96, line 33: Let's remove this when we update the base image since this will be an incorrect dockerfile. When you create an ubuntu release and update the versions of the other dockerfiles including this one we can remove it there.

PR #191, opened 2020-08-11
P.Session: changed mysql base to ubuntu-base

### LINKED ISSUE #190: P.Session: Use ubuntu-base for MySQL
Related to this issue 
https://github.com/CodeSignal/codesignal/issues/17888
To be consistent we should try to use ubuntu-base for MySQL image.

salbert96:
Closes #190 
This will change the base image for MySQL to the ubuntu base. Note that ubuntu base still has the old version of MySQL because the newer version breaks for DevOps and API tasks like this 
https://app.codesignal.com/admin-task/MEjf295sbz5Fv2kKp
Here is a discussion thread on slack https://codesignal.slack.com/archives/C0SJ40Z24/p1597149212022100

With the new base, we will be able to enable persistent sessions for MySQL.

Here are some proof of concepts 
Screen Shot 2020-08-12 at 15 58 15

Screen Shot 2020-08-11 at 19 59 20
Screen Shot 2020-08-11 at 19 59 51

Also, you can see the coderunner with version `709-change-mysql`
From here https://github.com/CodeSignal/coderunner/pull/710

**Note** that SQL version is change to 8.0.**21**

**Deprecated (This happened before the last commit about removing the old version)**
A strange thing happens, which I am not sure is because of my changes. The first time I click run I get this error.
SQL
And it is reproducible.


mysql/Dockerfile
```diff
@@ -45,12 +22,12 @@ RUN echo "mysql-apt-config	mysql-apt-config/select-server select mysql-8.0" | de
...
-    && chown -R mysql:mysql /var/log/mysql\
+    && chown -R mysql:mysql /var/log/mysql \
 # UTF-8 and bind-address
     && sed -i -e "$ a [client]\n\n[mysql]\n\n[mysqld]"  /etc/mysql/my.cnf \
     && sed -i -e "s/\(\[client\]\)/\1\ndefault-character-set = utf8/g" /etc/mysql/my.cnf \
     && sed -i -e "s/\(\[mysql\]\)/\1\ndefault-character-set = utf8/g" /etc/mysql/my.cnf \
     && sed -i -e "s/\(\[mysqld\]\)/\1\ninit_connect='SET NAMES utf8'\ncharacter-set-server = utf8\ncollation-server=utf8_unicode_ci\nbind-address = 0.0.0.0/g" /etc/mysql/my.cnf \
 # Clean up
-    && apt-get autoremove \
+    && apt-get autoremove -y \
```
salbert96, line 32: This was required since it deleted some packages
Screen Shot 2020-08-11 at 18 20 32
I assume the old packages from the old version.

mysql/Dockerfile
```diff
@@ -1,27 +1,6 @@
...
-ENV LANG='en_US.UTF-8' LANGUAGE='en_US:en' LC_ALL='en_US.UTF-8' DEBIAN_FRONTEND='noninteractive'
-
-#Install Node.js
-RUN apt-get update \
-    && curl -sL https://deb.nodesource.com/setup_8.x | bash - \
-    && apt-get install -y --no-install-recommends nodejs \
-    && apt-get -qq -y autoremove \
-    && apt-get autoclean \
-    && rm -rf /var/lib/apt/lists/*
+ENV DEBIAN_FRONTEND='noninteractive'
```
salbert96, line 3: This was the only difference of MySQL and ubuntu-base setup.

salbert96: It was an important one `¯\_(ツ)_/¯`

salbert96:
Here are some proof of concepts 
Screen Shot 2020-08-11 at 19 59 20
Screen Shot 2020-08-11 at 19 59 51

salbert96:
@menewman @Albinutte this does not work yet for some mysterious reason, but in general, are you okay with the approach of having MySQL 5 on ubuntu base and reinstall MySQL 8 here? I think MySQL 8 was problematic for DevOps and API tasks and that was the reason for having a separate SQL dockerfile for Database tasks.

menewman:
> in general, are you okay with the approach of having MySQL 5 on ubuntu base and reinstall MySQL 8 here?

My preference would be to have the same version of MySQL everywhere, and to have that version be the latest version everywhere. 😄 

If we already have MySQL 5 on base and MySQL 8 only for the MySQL image, I guess keeping that behavior is tolerable since it already exists, even if it's not ideal. Not sure if I understood that part right!

salbert96:
> Not sure if I understood that part right!

Yes, we have MySQL 5 on ubuntu-base and MySQL 8 on sql image which is not `FROM` ubuntu-base.
The goal is to make it from ubuntu-base while still not breaking anything. I also would prefer to have only one SQL version by only updating in ubuntu-base for example, but in general, I don't like even having database languages in ubuntu-base and MySQL image has some additional packages too. So the workaround was to just make it from ubuntu-base and re-install SQL 8 even though ubuntu-base has SQL 5.

menewman:
If that's already what we have, and this doesn't make it worse, then I'm okay with it. 🙂 

> but in general, I don't like even having database languages in ubuntu-base

The reason why we do this is because some questions need to interact with databases via a driver in another language. For example, I might want to ask a candidate to write a Python script to interact with a MySQL database, rather than having them write a MySQL query directly. In that case, we need to have MySQL available in the Python image, or vice versa. Since we have many coding languages and comparatively few database languages, simply installing the databases in the base image seemed like a valid way to handle that use case.

We may need to reconsider that as we add more database variants.

> MySQL image has some additional packages too

That part is fine, and doesn't necessarily impact the versioning question, since the packages already work with the more current MySQL version

PR #192, opened 2020-08-12
update ubuntu base dependencies

lizhoucheng:
update the dependencies to use base v7.1 which is successfully built


salbert96:
Can we also update mysql?

Albinutte:
Great catch @salbert96! @lizhoucheng if you include mySQL in this update, please don't forget to update build rule at https://hub.docker.com/repository/docker/codesignal/mysql/builds to make sure a new tag will be built when you create the `based-on/ubuntu-base/v7.1` tag.

salbert96:
Ok, never mind about mysql, I will update it myself.

Issue #93, opened 2019-11-07
Install R package tryCatchLog from the CRAN repository


menewman:
Previously, we had to switch to installing `tryCatchLog` from source, because it was removed from CRAN due to a failing test:
https://github.com/CodeSignal/dockerfiles/pull/92#discussion_r341745899

However, the maintainer has fixed the issue and republished the package:
https://github.com/aryoda/tryCatchLog/issues/47#issuecomment-550965382

As a result, although it is not urgent, we should switch back to installing from CRAN whenever we next build this image.

Safe to hold this until the next time we need to modify the R Dockerfile.

PR #201, opened 2020-09-21
add matrix-stats and gmcm

lizhoucheng:
close https://github.com/codesignal/coderunner/issues/772 and #93 

Add matrix-stats and gmcm packages and switch back to the old way to install tryCatchLog. It builds successfully on Dockerhub.


menewman:
@lizhoucheng - Not mandatory but you could also fix this bonus issue while you're in the R Dockerfile maybe? 🙏 

https://github.com/CodeSignal/dockerfiles/issues/93

lizhoucheng:
Sure!

ivyc81 (changes requested):
Looks like GMCM is not installed correctly?
This is from dockerhub(https://hub.docker.com/repository/registry-1.docker.io/codesignal/r-env/builds/9bb06b2b-2eed-4b8c-9011-daf070f0b273)
Screen Shot 2020-09-23 at 8 45 19 AM

I pulled the image and played with it on my local
Screen Shot 2020-09-23 at 8 47 27 AM

lizhoucheng:
Good catch! I install `r-base-dev` which we should install since we are using `install.packages()`. (ref: https://cran.r-project.org/bin/linux/ubuntu/README.html). It fixes the problem

ivyc81 (approved):
🎉 Good job finding the missing puzzle to install GMCM!

PR #208, opened 2020-10-23
Base image version updates: Node to 14, MongoDB to 4.4, PostgreSQL to 13

menewman:
Updates these versions in our base image to latest:
- Node to 14.14.0
- MongoDB to 4.4.1
- PostgreSQL to 13

This only updates the base Dockerfile; follow-ups will be necessary to update its descendants, as usual.


base/Dockerfile
```diff
@@ -69,12 +62,13 @@ RUN echo "deb http://cn.archive.ubuntu.com/ubuntu/ xenial main restricted univer
...
 
 # Install mongodb server
-RUN apt-get update \
-    && apt-get install -y mongodb-server --no-install-recommends \
-    && apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 9DA31620334BD75D9DCB49F368818C72E52529D4 \
-    && echo "deb [ arch=amd64,arm64 ] https://repo.mongodb.org/apt/ubuntu xenial/mongodb-org/4.0 multiverse" | tee /etc/apt/sources.list.d/mongodb-org-4.0.list \
+RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 20691EEC35216C63CAF66CE1656408E390CFB1F5 \
+    && echo "deb [ arch=amd64,arm64 ] https://repo.mongodb.org/apt/ubuntu xenial/mongodb-org/4.4 multiverse" | tee /etc/apt/sources.list.d/mongodb-org-4.4.list \
     && apt-get update \
-    && apt-get install -y mongodb-org=4.0.5 mongodb-org-server=4.0.5 mongodb-org-shell=4.0.5 mongodb-org-mongos=4.0.5 mongodb-org-tools=4.0.5 \
```
menewman, line 77: If you're wondering why I shortened the installation command:

`mongodb-org` is a metapackage that contains all the other packages, so we can simply install `mongodb-org` without listing out every subpackage specifically.

base/Dockerfile
```diff
@@ -106,17 +100,13 @@ RUN echo "deb http://apt.postgresql.org/pub/repos/apt/ xenial-pgdg main" >> /etc
...
-       postgresql-contrib-12 \
+       postgresql-13 \
+       postgresql-client-13 \
+       postgresql-contrib-13 \
        postgis \
-       postgresql-12-postgis-2.5 \
+       postgresql-13-postgis-2.5 \
     && rm -rf /var/lib/apt/lists/*
 
-# Installing Node 10.15.3
```
menewman, line 116: This Node version was previously necessary for an older session server implementation, but it's not being used anymore and hasn't for some time!

base/Dockerfile
```diff
@@ -69,12 +62,13 @@ RUN echo "deb http://cn.archive.ubuntu.com/ubuntu/ xenial main restricted univer
     && rm -rf /var/lib/apt/lists/*
 
 # Install mongodb server
-RUN apt-get update \
-    && apt-get install -y mongodb-server --no-install-recommends \
```
lizhoucheng, line 73: Just want to make sure do we still need this?

menewman: Nah, I don't think we ever needed it, and it was probably originally added as a hack to work around some kind of configuration issue (like the one I suppressed here with a symlink).

Ubuntu includes packages `mongodb`, `mongodb-server`, etc. by default, but these don't contain the most up-to-date version of Mongo. So, instead, the recommended way is to use `mongodb-org` and its sub-packages, and it's not recommended that you use both together because it can lead to conflicts and errors.

menewman:
@lizhoucheng - Bump!

lizhoucheng (approved):

PR #212, opened 2020-11-02
Updating Kotlin 

ArDrambyan:
## Changes
Updated Kotlin version to 1.4.10, no need to install coroutines because the lib can be found in `/opt/kotlinc/lib`


menewman:
@ArDrambyan - Thanks for doing this! Just want to note that I'm in the process of updating image versions based on changes to the "base image," so please ping me before releasing a new Kotlin version after this merges? Otherwise we might have some image version number "conflicts."

menewman (approved):

menewman:
I'm actually going to go ahead and merge this so I can capture it in the next Kotlin release -- sorry if you meant to make more changes! Looks ready to me though

PR #218, opened 2020-11-22
Switch Java docker image version in all JVM languages

roma0297:
https://github.com/CodeSignal/coderunner/issues/882


Albinutte (approved):
👍

Issue #206, opened 2020-10-20
Update Groovy to v3


lizhoucheng:
A follower up for https://github.com/CodeSignal/coderunner/issues/657

Right now, the Groovy compiler throw the message:
```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.codehaus.groovy.reflection.CachedClass (file:/opt/groovy/lib/groovy-2.4.15.jar) to method java.lang.Object.finalize()
WARNING: Please consider reporting this to the maintainers of org.codehaus.groovy.reflection.CachedClass
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations\nWARNING: All illegal access operations will be denied in a future release
```

This message is filtered out by its language template handler. Still, it is better to solve this. There are some related info like this report: https://issues.apache.org/jira/browse/GROOVY-9070

But the preferred solution is to update groovy version (not sure whether the newer version will make the warnings go away). We're using `v2.4.15`, while version 4 (in alpha stage) is already [available](http://groovy-lang.org/download.html)!

menewman:
We can try updating to Groovy. But, we should update to latest stable instead of bleeding edge alpha (i.e., update to 3+ instead of 4+).

menewman:
@lizhoucheng - Can you help me understand why updating the Groovy version would fix this? Isn't this an error in the template code or in an external dependency?

lizhoucheng:
I am also not sure whether it helps. I read through some issue reports and they seem related to some specific version of Groovy, or JDK. I think the person who takes this issue need to figure out whether updating the version helps or not. If not, then some hacks need to be used to repress the warning

menewman:
Ah, if you look at this duplicate JIRA issue, it's tagged as being fixed in 3.0.0-beta-1:
https://issues.apache.org/jira/browse/GROOVY-8339

So, hopefully you're right that updating the version will resolve the warning.

PR #219, opened 2020-11-24
Groovy update to 3.0.6

lizhoucheng:
First part of https://github.com/CodeSignal/dockerfiles/issues/206

Update Groovy to 3.0.6.  

Docker build: https://hub.docker.com/repository/registry-1.docker.io/codesignal/groovy/builds/95ff5422-dd7b-42bf-b70a-9405339ffc46

Coderunner that points to the new image: http://157.230.180.133:12345/


ivyc81 (approved):
🎉  Code change looks good.

One question tho, why are you using a manual build with tag 206-groovy instead of just targeting tag dev in your coderunner change. Is there something wrong with dev build?

lizhoucheng:
Notthing wrong! For some reason, when I make the coderunner PR, the labels are not loaded for some reasons, maybe it was related to my internet speed or browser. So I went on to do the traditional way :). But having the dev tag is definitely easier!

PR #217, opened 2020-11-22
Add libjackson2-dataformat-yaml lib to Java docker image

roma0297:
https://github.com/CodeSignal/coderunner/issues/882

PR #229, opened 2021-01-10
Replace kotlinc with gradle script

roma0297:
#217 

[Asana original task link](https://app.asana.com/0/417895759505117/1195673034468681/f)

The idea is to replace `kotlinc` with gradle script because `kotlinc` doesn't support asterisk in the classpath. It also makes dependencies management much easier.

In order to test that everything works fine, you can just try to run multiple tasks in kotlin with classes from different libraries like jackson etc.


kotlin/build.gradle
```diff
@@ -0,0 +1,23 @@
...
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    implementation fileTree(dir: '/usr/share/java', include: '*.jar')
+    implementation platform('org.jetbrains.kotlin:kotlin-bom')
+    implementation 'org.jetbrains.kotlin:kotlin-stdlib-jdk8'
+    implementation 'com.fasterxml.jackson.module:jackson-module-kotlin:2.12.1'
```
DimaPhil, line 14: I see that `jackson-dataformat-yaml` version is `2.11.3` in Java Dockerfile right now, is it ok? Can't we make this version shared between gradle files to avoid inconsistencies?

kotlin/build.gradle
```diff
@@ -0,0 +1,23 @@
+plugins {
+    id 'org.jetbrains.kotlin.jvm' version '1.4.21'
```
DimaPhil, line 2: In Dockerfile, Kotlin version is 1.4.10, and here it is 1.4.21 - is it ok?

kotlin/Dockerfile
```diff
@@ -10,3 +10,8 @@ RUN KOTLIN_VERSION=1.4.10 \
   && rm -rf /var/lib/apt/lists/*
 
 ENV PATH $PATH:/opt/kotlinc/bin
+
+# upload build.gradle script
+COPY build.gradle /usercode/build.gradle
+RUN cd /usercode \
```
Albinutte, line 16: Does this use the fact that user code will be executed in the `/usercode` directory? If yes, I strongly suggest changing this; it is not guaranteed that the directory will remain the same in the future, and if someone changes the directory in the `coderunner` repo, they will have no idea that kotlin dockerfile will break. Is it possible to build the libraries somewhere in the shared folder, or specify a separate libraries path to kotlin when executing user code?

Albinutte (changes requested):
I agree with Dima's comments about libraries version; perhaps we could reuse shared resource file to make sure the libraries are synced?

Another point: could you update PR description with the link to the actual issue (is it #882?), and some basic testing instructions? Given that I pull the image locally, how can I make sure that the required libraries are installed?

menewman:
I am removing my review request since I don't have anything to add that Albina and Dima didn't already point out -- I agree with their comments, though!

kotlin/build.gradle
```diff
@@ -0,0 +1,23 @@
+plugins {
+    id 'org.jetbrains.kotlin.jvm' version '1.4.10'
```
DimaPhil, line 2: Don't we need to move this version to env variables as well?

DimaPhil (approved):

kotlin/Dockerfile
```diff
@@ -10,3 +10,8 @@ RUN KOTLIN_VERSION=1.4.10 \
   && rm -rf /var/lib/apt/lists/*
 
 ENV PATH $PATH:/opt/kotlinc/bin
+
+# upload build.gradle script
+COPY build.gradle /usr/share/java/build.gradle
+RUN cd /usr/share/java \
+    && gradle build
```
Albinutte, line 17: Would be nice to remove `build.grade` after everything's installed, just to make the layer slightly smaller.

Albinutte (commented):
I assume #232 changes need to be downloaded first, so I'm not approving just yet.

In the mean time, I'd still like to get PR description updated 🙂 
1. For posterity, please mention the issue you're trying to close in the PR description.
2. To help me make sure the issue is resolved, please add some testing instructions as well.

kotlin/Dockerfile
roma0297: Nope, we're gonna need this build gradle to run the code

kotlin/build.gradle
roma0297: I don't think so, because it's only used in 1 place. For Jackson it was important because we use it in multiple places and the version should be in sync

DimaPhil: As far as I know, this version is also used in Kotlin Dockerfile itself, isn't it?

kotlin/Dockerfile
Albinutte: 1. Isn't it a bit slow, running gradle build` on every run request? How long would it take?
2. Does `gradle build` require network access? If it does, then it won't be possible to run it for every task -- some tasks don't have network access enabled.

DimaPhil: @Albinutte Let me quickly explain why are we going to use gradle and why it should be ok in terms of efficiency and other stuff. Currently Java and Kotlin are being compiled just using CLI compiler (javac / kotlinc), which is not efficient for libraries management - you can only provide the list of JARs and don't have any power about anything but the order of libraries (in case two different JARs include the same dependency but with different versions, we'll have both of them in the classpath, which causes JAR hell and definitely something we have to avoid).

We're going to switch to gradle, which is a dependency management tool for Java and run all tasks using it instead. On image build, gradle will download all the needed dependencies and store them in a separate folder. On task run, gradle will just include all the JARs from this folder and provide them to the classpath - so there is no other dependency downloading here; gradle is used just as a wrapper for compiler here

Albinutte: Awesome, thanks for the explanation @DimaPhil! It all makes sense then 👍 

I'd suggest leaving a comment in this file with a similar explanation; to an ignorant reader (hello!) this might look like a leftover artifact 🙂  It's especially confusing because we _do_ delete `build.gradle` in Java dockerfile.

DimaPhil: @Albinutte Makes sense! @roma0297 Can you please add the comment Albina asks for?

Regarding Java btw - I see why it's confusing, from what I see the main difference between Java and Kotlin is that Java allows to use wildcards in classpath (so you can just include the whole folder of jar to it), while Kotlin doesn't allow it. Because of that in Java we don't need gradle to run the code - we can simple use `java -cp <classpath>` and provide the JARs folder - gradle already did all the dependency management job for us. But for Kotlin we can't do that, so we use gradle as a wrapper to automatically provide every single JAR from the folder to the classpath

kotlin/Dockerfile
roma0297: Done :)

kotlin/build.gradle
roma0297: Not sure whether we need Kotlin CLI at all, let my try it out without it

kotlin/Dockerfile
roma0297: @Albinutte sure, it's a good point, I just added a few comments to the Kotlin Dockerfile. I also removed Kotlin CLI at all because we can move to gradle script completely

Albinutte (approved):
👍 Thanks for bearing with me! I'm going to copy-paste my previous high-level comment though 🙂 

1. For posterity, please mention the issue you're trying to close in the PR description.
2. To help me make sure the issue is resolved, please add some testing instructions as well.

PR #277, opened 2021-01-21
Updating ubuntu version for base

### LINKED ISSUE #234: Update ubuntu-base image to support new Ubuntu version

ArDrambyan:
Closes #234 

Updating ubuntu version to 20.04 for base images


ArDrambyan:
@menewman - I deployed coderunner with a new docker image and all tests passed, it is deployed in the staging rundash 157.245.88.31. 

I updated `276-ubuntu-version-20` branch to build dockers with tag `new` instead of `dev` to be able to develop separately

menewman:
I didn't have time to review this carefully today -- I'm sorry! Hope to get to it tomorrow 🙏

base/Dockerfile
```diff
@@ -1,25 +1,32 @@
 ###
 # A base Docker Container, build on ubuntu and includes node.js, mysql, mongodb, and git.
 ###
-FROM ubuntu:16.04
+FROM ubuntu:20.04
+
+# Making tzdata installion noninteractive
```
menewman, line 6: [trivial]
```suggestion
# Making tzdata installation noninteractive
```

base/Dockerfile
```diff
@@ -1,25 +1,32 @@
...
+ENV DEBIAN_FRONTEND noninteractive
+ENV DEBCONF_NONINTERACTIVE_SEEN true
+RUN echo 'tzdata tzdata/Areas select Etc' | debconf-set-selections \
```
menewman, line 9: Hm, for my own knowledge, what are these two new commands for? Does one of the other packages further down not work correctly if we don't set these selections first?

base/Dockerfile
```diff
@@ -1,25 +1,32 @@
...
+        bc \
+        locales \
+        build-essential \
+        curl \
+        wget \
+        apt-transport-https \
+        apt-utils \
+        xz-utils \
+        gnupg \
+        ca-certificates \
```
menewman, line 25: Looks like we swapped software-properties-common with ca-certificates? Curious, what prompted the change here? (I'm sure there's a reason and would love to learn)

EDIT: I see we're adding software-properties-common below when installing git

base/Dockerfile
```diff
@@ -62,8 +69,8 @@ RUN echo "deb http://cn.archive.ubuntu.com/ubuntu/ xenial main restricted univer
     && rm -rf /var/lib/apt/lists/*
 
 # Install mongodb server
-RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 20691EEC35216C63CAF66CE1656408E390CFB1F5 \
-    && echo "deb [ arch=amd64,arm64 ] https://repo.mongodb.org/apt/ubuntu xenial/mongodb-org/4.4 multiverse" | tee /etc/apt/sources.list.d/mongodb-org-4.4.list \
+RUN curl -fsSL https://www.mongodb.org/static/pgp/server-4.4.asc | apt-key add - \
+    && echo "deb [ arch=amd64,arm64 ] https://repo.mongodb.org/apt/ubuntu focal/mongodb-org/4.4 multiverse" | tee /etc/apt/sources.list.d/mongodb-org-4.4.list \
     && apt-get update \
     # mongodb-org-server tries to call systemctl, which we don't have, so symlink it to just return "true"
```
menewman, line 75: This hack might not be necessary anymore now that we're on a higher Ubuntu version -- might be worth trying to remove the `ln` call (and the matching `rm` for the fake `systemctl` file below) to see if everything is okay?

We may also be able to start the Mongo service differently without having to manually copy the service file after upgrading, which would be nice.

base/Dockerfile
```diff
@@ -92,21 +99,23 @@ RUN apt-get update \
...
-       postgis \
-       postgresql-13-postgis-2.5 \
+        postgresql-13 \
+        postgresql-client-13 \
+        postgresql-contrib-13 \
+        postgis \
+        postgresql-13-postgis-2.5 \
     && rm -rf /var/lib/apt/lists/*
 
 # Installing PowerShell
```
menewman, line 121: The 16.04 in the URL below seems a little concerning 😅

menewman (commented):
This looks good and it's great to see the MySQL & PostgreSQL tests passing on the deployed box!

I just left a couple of really small questions, and an observation about Mongo installation & service start. In addition to that, I wanted to point out that this wouldn't be a bad time to update a couple of minor/patch versions in the base image while we're here. The following things all have a minor/patch version bump ready, which should hopefully be safe to do in-place:
- Node
- MySQL
- PostgreSQL
- Mongo

Not going to strictly block on that because it's kind of a scope-creepy "while you're here" request (not good practice for me as a reviewer). But, if you're willing, it's appreciated, since version bumps from the base image can kind of be a pain and are nice to batch together 😄

base/Dockerfile
ArDrambyan: tzdata installation is interactive, so with environment variables, I'm making sure that it will be done in non-interactive mode and that `echo ...` adds all necessary configuration for tzdata installation

base/Dockerfile
ArDrambyan: I didn't swap them, we need ca-certificates for MongoDB installation and installing `software-properties-common` before MongoDB raised some problems so I'm moved if after MongoDB installation

ArDrambyan:
@menewman - Done 😅  you can check compiler versions here: http://192.81.211.204:12345, they all are up-to-date

base/Dockerfile
menewman: Right, I guess what I'm asking is which installation below requires that we set up `tzdata` first, since it seems like we didn't bother doing this in 16.04?

menewman (approved):
:shipit: 👌 

Thanks for updating the versions!

ArDrambyan:
Before merging I want to ask you to take a look at this confluence doc for updating docker images: https://codesignal.atlassian.net/wiki/spaces/PE/pages/945946690/How+to+update+docker+images+to+Ubuntu+20.04

menewman:
@ArDrambyan - Thanks for writing it up! I don't have any blocking comments on that article, so feel free to merge 🚢

PR #279, opened 2021-01-29
[Content] Support express for NodeJS

Anastasiya-Zhyrkevich:


js/Dockerfile
```diff
@@ -5,10 +5,12 @@ RUN npm install -g --unsafe-perm=true \
     async \
     axios \
     chai \
+    chai-http \
     co \
     connect \
     crypto-js \
     debug \
+    express \
```
Albinutte, line 13: If we're installing these libraries for JS, would it also make sense to install `@types` libraries for TypeScript (in [here](https://github.com/CodeSignal/dockerfiles/blob/master/typescript/Dockerfile))? Without the types, it will not be possible to use these libraries with TypeScript.

Anastasiya-Zhyrkevich: Added 
According to the investigation - https://www.npmjs.com/package/@types/chai-http is deprecated, no need for it!

Albinutte (approved):
Great! 👍

PR #282, opened 2021-02-03
Python Ubuntu 20.04

### LINKED ISSUE #264: Update python2 image to support new Ubuntu version

ArDrambyan:
Closes #264 #265 

## Changes
Updated `Python2` and `Python3` images to be based on `Ubuntu 20.04`


ArDrambyan:
@menewman - I don't have any fully tested coderunner for now, but I deployed coderunners 68.183.28.148 and 142.93.58.135 with these changes, so the tests should be finished soon

python2/Dockerfile
```diff
@@ -11,6 +11,8 @@ RUN apt-get update \
     liblzma-dev \
   && rm -rf /var/lib/apt/lists/*
 
+RUN wget --quiet https://raw.githubusercontent.com/paulfitz/mysql-connector-c/master/include/my_config.h -P /usr/include/mysql/
```
menewman, line 14: Reliance on the `master` branch of someone's GitHub repo that we don't have control over makes me a little nervous. Why do we need this for the new Ubuntu version when we didn't before?

menewman (changes requested):
Although the tests are not done, it looks like there may be an issue with Python 3 raw tests! Template tests are still going, so I'm not sure about those ones.

Screen Shot 2021-02-03 at 1 20 08 PM

python2/Dockerfile
ArDrambyan: I was getting `my_config.h` missing error for `MySql-Python` package installation. I was trying to fix it by installing some package instead of copying that file but I couldn't find any package that will fix this issue. I Will look into it one more time

ArDrambyan: I downloaded `my_config.h` from https://dev.mysql.com/doc/dev/mysql-server/8.0.11/my__config_8h_source.html to remove any dependency

ArDrambyan:
@menewman - finally coderunner 157.230.212.197 with latest changes passes the tests

menewman (approved):
Thanks for the changes, Aram! I am still a little worried about the config file (I think it must be masking another issue, right?) but willing to approve -- my concern about depending on an unreliable dependency is resolved by at least copying the file into this repo 👍

ArDrambyan:
> I am still a little worried about the config file (I think it must be masking another issue, right?)

I did a lot of search for it and looks like it's not a rare issue and the main solution in all answers was something like - "make sure that my_config.h file exists in /usr/lib/mysql directory" and I don't know why we didn't face this issue on the ubuntu 16.04

Issue #169, opened 2020-06-11
codesignal/acceptance-tests latest versions are not working


salbert96:
The latest versions of `codesignal/acceptance-tests` are failing coderunner tests.

We're using an older version of this image (v5.5) that does pass the tests, similar to what we do for the Objective-C image.

menewman:
@salbert96 - You mean, after updating the versions from your latest PRs? The tests were passing on the latest deployed version of the images.

menewman:
We're going to assume the former and send this directly to you, but please clarify if we misunderstood!

salbert96:
The image was not updated the last time ubuntu-base was updated.
The version is `codesignal/acceptance-tests:v5.5` in coderunner, but the last ubuntu base change was `v5.6` I am not sure if it would pass with version 5.6 but I did not go into the details of that image.

I would prefer to move it to the icebox for now, since it would be better to do this after persistent sessions.

salbert96:
@menewman ^^

menewman:
I see! Removing your assignment then

ArDrambyan:
Resolved in the sope of #288

PR #288, opened 2021-02-08
acceptance-tests Ubunutu 20.04

### LINKED ISSUE #235: Update acceptance-tests image to support new Ubuntu version

ArDrambyan:
Closes #235 

## Changes

Updated base image for `acceptance-tests` to support Ubuntu 20.04


ArDrambyan:
coderunner `192.241.183.49 ` is deployed but the `puppeteer` test is failing and I can't figure out why

menewman (commented):
@ArDrambyan - Hm, I would guess that the issue is the pre-existing problem described here (seems that we haven't addressed it):
https://github.com/CodeSignal/dockerfiles/issues/169

Not sure how to review this one. I do agree it's _probably_ not something new that resulted from the base image change. On the other hand, that prior issue makes it hard to tell whether everything has gone well with the acceptance-tests image bump.

ArDrambyan:
@menewman - after some changes you can see that coderunners `162.243.27.174` and `162.243.27.170 ` are passing all the tests 🥳 
I rewrote a bit dependency installation for `puppeteer` but it looks like the main reason was `NODE_PATH` overriding, it was incorrect, and importing `puppeteer` was throwing an exception.

acceptance-tests/Dockerfile
```diff
@@ -1,18 +1,35 @@
...
+FROM codesignal/ubuntu-base:new
 
-# Install latest chrome dev package and fonts to support major charsets (Chinese, Japanese, Arabic, Hebrew, Thai and a few others)
-# Note: this installs the necessary libs to make the bundled version of Chromium that Puppeteer installs, work.
-RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \
-    && sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list' \
-    && apt-get update \
-    && apt-get install -y google-chrome-unstable fonts-ipafont-gothic fonts-wqy-zenhei fonts-thai-tlwg fonts-kacst ttf-freefont \
-      --no-install-recommends \
+# Install chromium browser, necessary libraries for puppeteer and fonts to support major charsets (Chinese, Japanese, Arabic, Hebrew, Thai and a few others)
```
menewman, line 3: Does this conflict at all with the comment on line 30? i.e., this says "install chromium browser" but below we say it's not necessary because it already comes with Puppeteer

acceptance-tests/Dockerfile
```diff
@@ -1,18 +1,35 @@
...
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
```
menewman, line 5: Curious, was there documentation or a reference that helped you figure out that we need to install these particular libs? Might be nice to add in a comment so that it's easy to check later

menewman (approved):
Awesome; you really rock for figuring this out! 🚀 

I have a couple of small questions I wanted to ask (not exactly blockers, but would like to know, esp. about the new dependencies)

acceptance-tests/Dockerfile
ArDrambyan: Nice catch! I was installing chromium explicitly but then I noticed that the `puppeteer` installs it anyway so I removed the installation but forgot to update the comment 😅

acceptance-tests/Dockerfile
ArDrambyan: I used some article for this list, I'll add the link to it in the comment

PR #290, opened 2021-02-08
Install source-map-support in TS image

Albinutte:
Adds [`source-map-support`](https://www.npmjs.com/package/source-map-support) to TypeScript dockerfile to enable printing out correct error positions. It is a supporting PR for https://github.com/CodeSignal/coderunner/issues/989; I will submit the main PR in the coderunner repo shortly.

I briefly considered installing it for CoffeScript as well, but there's no need to -- CoffeeScript already works just fine.


menewman (approved):
:shipit:

PR #299, opened 2021-02-13
GCC 10.x

ArDrambyan:
## Changes

I forgot to update the GCC version while updating the Ubuntu version for C++, so I'm doing it with separate PR.

You can see that coderunner [104.248.57.138](https://rundash-staging.codesignal.com/coderunner-test/104.248.57.138/test-results) is passing all the tests and the compiler version for GCC is now 10.2.0

Also, I committed some changes to coderunner's `testing-ubuntu-20` to make everything work (updated flag to support c++20 and replaced g++ with of g++-8)


menewman (approved):
Very exciting!

PR #302, opened 2021-02-18
Ruby Ubuntu 20.04

### LINKED ISSUE #267: Update ruby image to support new Ubuntu version

ArDrambyan:
Closes #267 

## Changes

Updated `ruby` docker image to derive from `codesignal/ubuntu-base:new` image to be based on Ubuntu 20.04

Coderunner 206.189.185.112 contains the latest changes and passes the tests


menewman (approved):
Looks good!

I was also checking for a new Ruby version in case we could update the minor/patch version, but it looks like the next one is 3.0, and that one probably deserves to be a separate PR/project since there might be breaking changes.

ArDrambyan:
> and that one probably deserves to be a separate PR/project since there might be breaking changes.

I created an issue for that - #304

menewman:
Ah, thanks for writing that up!

PR #309, opened 2021-03-01
OpenCV Ubuntu 20.04

### LINKED ISSUE #237: Update opencv image to support new Ubuntu version

ArDrambyan:
Closes #237 

## Changes
Updated base branch for `OpenCV` to be derived from `Ubuntu 20.04` and updated version to 4.5.1 (latest)


ArDrambyan:
Coderunners https://rundash-staging.codesignal.com/coderunner-test/159.89.236.30/test-results and https://rundash-staging.codesignal.com/coderunner-test/142.93.250.114/test-results are passing the tests

opencv/Dockerfile
```diff
@@ -15,7 +15,6 @@ RUN add-apt-repository -y ppa:ubuntu-toolchain-r/test \
     libswscale-dev \
     unzip \
 # OpenCV optional dependencies
-    python-dev \
```
menewman, line 18: Why remove this one in particular?

menewman (approved):
Ah, great. This builds and everything!

opencv/Dockerfile
ArDrambyan: because its installation was failing 😅

menewman: This doesn't make me feel great 😄 

The tough thing is, since this is listed as an "optional dependency" but without a link to install docs that explain why we'd want it, it's hard to understand what impact removing this dependency will have. 

What about replacing `python-dev` and `python-numpy` with `python3-dev python3-numpy` -- might it help?

ArDrambyan: I think that package was needed for the build but the build is passing successfully  and `python3-*` packages won't help because the box has only python2 by default

menewman: > I think that package was needed for the build but the build is passing successfully

If the build passes, then I think something doesn't add up there...? I guess mostly likely these packages would be used for Python bindings to the OpenCV library, rather than for building/compiling it? In that case, maybe that doesn't matter if we generally intend for OpenCV to be interacted with via C++ instead.

PR #315, opened 2021-03-03
PHP add 2 packages

Anastasiya-Zhyrkevich:
Added packages: 
- symfony/process
- guzzlehttp/guzzle

### DIFF
--- a/php/Dockerfile
+++ b/php/Dockerfile
@@ -10,7 +10,19 @@ RUN add-apt-repository ppa:ondrej/php \
     php8.0-xml \
   && rm -rf /var/lib/apt/lists/*
 
-# Install PHPUnit
-RUN wget -O phpunit https://phar.phpunit.de/phpunit-9.phar \
-  && chmod +x phpunit \
-  && mv phpunit /usr/local/bin/phpunit
+# Install composer 
+RUN php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');" \
+  && php -r "if (hash_file('sha384', 'composer-setup.php') === '756890a4488ce9024fc62c56153228907f1545c228516cbf63f885e036d37e9a59d27d63f46af1d4d07ee0f76181c7d3') { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;" \
+  && php composer-setup.php --filename=composer \
+  && php -r "unlink('composer-setup.php');"
+
+# Install required packages
+RUN ./composer global require phpunit/phpunit=9.3 \
+  && ./composer global require symfony/process=3.3 \
+  && ./composer global require guzzlehttp/guzzle=6.3 
+
+# Expose phpunit to PATH and clean up. COMPOSER_HOME_PATH - is a directory, where composer installed the packages
+RUN COMPOSER_HOME_PATH=$(./composer global config bin-dir --absolute) \
+  && sudo ln -s ${COMPOSER_HOME_PATH}/phpunit phpunit \
+  && export PATH=${COMPOSER_HOME_PATH}:$PATH
+  && rm -rf composer


Anastasiya-Zhyrkevich:
Is it fine that build time is around 7 mins? 
Is it much more than it was before?
image

php/Dockerfile
```diff
@@ -8,9 +8,22 @@ RUN add-apt-repository ppa:ondrej/php \
     php8.0-curl \
     php8.0-mysql \
     php8.0-xml \
+    php8.0-mbstring \
```
Anastasiya-Zhyrkevich, line 11: I have updated `phpunit` for 9.3 version, this extention is required.

php/Dockerfile
```diff
@@ -8,9 +8,22 @@ RUN add-apt-repository ppa:ondrej/php \
...
   && rm -rf /var/lib/apt/lists/*
 
-# Install PHPUnit
-RUN wget -O phpunit https://phar.phpunit.de/phpunit-9.phar \
-  && chmod +x phpunit \
-  && mv phpunit /usr/local/bin/phpunit
+# Install composer
```
Anastasiya-Zhyrkevich, line 14: I decided to use package manager for PHP, as I failed using `apt-get` for my needs

I wanted to install `php-symfony-process` and `guzzlehttp`, but my project was refusing to see the packages installed. 

https://github.com/Anastasiya-Zhyrkevich/phpAPIConstructionExample

php/Dockerfile
```diff
@@ -8,9 +8,22 @@ RUN add-apt-repository ppa:ondrej/php \
...
+  && php -r "unlink('composer-setup.php');"
+
+# Install required packages
+RUN ./composer global require phpunit/phpunit=9.3 \
+  && ./composer global require symfony/process=3.3 \
+  && ./composer global require guzzlehttp/guzzle=6.3 
+
+# Expose phpunit to PATH and clean up. COMPOSER_HOME_PATH - is a directory, where composer installed the packages
+RUN COMPOSER_HOME_PATH=$(./composer global config bin-dir --absolute) \
+  && ln -s ${COMPOSER_HOME_PATH}/phpunit phpunit \
```
Anastasiya-Zhyrkevich, line 27: Doing this to be able to run `phpunit tests --test-suffix=.php`, as this is a COMAND line for PHP filesystem tasks.

php/Dockerfile
```diff
@@ -8,9 +8,22 @@ RUN add-apt-repository ppa:ondrej/php \
...
     php8.0-xml \
+    php8.0-mbstring \
   && rm -rf /var/lib/apt/lists/*
 
-# Install PHPUnit
-RUN wget -O phpunit https://phar.phpunit.de/phpunit-9.phar \
-  && chmod +x phpunit \
-  && mv phpunit /usr/local/bin/phpunit
+# Install composer 
+RUN php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');" \
```
Anastasiya-Zhyrkevich, line 15: From official page https://getcomposer.org/download/

ArDrambyan:
> Is it fine that build time is around 7 mins?
> Is it much more than it was before?
> image

I think yes, you changed a bit installation rules and added new packages so it can take a bit longer and 7 minutes is not critical

ArDrambyan:
@Anastasiya-Zhyrkevich can you please deploy a coderunner with these changes? (In case if you already didn't)

If you are not familiar with deploying coderunner to test changes:
1. Create a new branch in the coderunner (something like `php-new-packages` or <issue-number>-php-new-packages if an issue for this task exists), you can even re-use the same branch to update the image tag after merging this PR and drafting a new release.
2. Update `PHP`'s image tag to `dev` in the `compilers.ts` and `UpdateDocker.sh` files.
3. This is one optional but I'll recommend adding(modifying) tests for PHP to also test these libs

php/Dockerfile
```diff
@@ -8,9 +8,22 @@ RUN add-apt-repository ppa:ondrej/php \
...
+
+# Install required packages
+RUN ./composer global require phpunit/phpunit=9.3 \
+  && ./composer global require symfony/process=3.3 \
+  && ./composer global require guzzlehttp/guzzle=6.3 
+
+# Expose phpunit to PATH and clean up. COMPOSER_HOME_PATH - is a directory, where composer installed the packages
+RUN COMPOSER_HOME_PATH=$(./composer global config bin-dir --absolute) \
+  && ln -s ${COMPOSER_HOME_PATH}/phpunit phpunit \
+  && export PATH=${COMPOSER_HOME_PATH}:$PATH \
```
ArDrambyan, line 28: please use
`ENV PATH ${COMPOSER_HOME_PATH}:$PATH` instead of `RUN export ...`, `ENV` guarantees that the environment variable changes will be saved for the container. More about it you can read from here, for example: https://stackoverflow.com/questions/33379393/docker-env-vs-run-export

php/Dockerfile
ArDrambyan: Can you please add the same comment (or even a simplified one) directly in the dockerfile? It can be really helpful in the future!

ArDrambyan (changes requested):
Great Job! I love the idea of using the PHP package manager, thank you these changes 🥇 
I just left some small comments.

p.s. not approving until the coderunner is deployed

Anastasiya-Zhyrkevich:
Staging 134.209.122.138

php/Dockerfile
```diff
@@ -8,9 +8,23 @@ RUN add-apt-repository ppa:ondrej/php \
...
+# Install required packages
+RUN ./composer global require phpunit/phpunit=9.3 \
+  && ./composer global require symfony/process=3.3 \
+  && ./composer global require guzzlehttp/guzzle=6.3 
+
+# Expose phpunit to PATH and clean up. COMPOSER_HOME_PATH - is a directory, where composer installed the packages
+# Adding symbolic link to run filesystem tests `phpunit tests --test-suffix=.php`
+RUN COMPOSER_HOME_PATH=$(./composer global config bin-dir --absolute) \
+  && ln -s ${COMPOSER_HOME_PATH}/phpunit phpunit \
+  && ENV PATH ${COMPOSER_HOME_PATH}:$PATH \
```
ArDrambyan, line 29: `ENV` is a dockerfile command like `RUN` so it should separate from the `RUN` command, that's why the build is failing

ArDrambyan:
@Anastasiya-Zhyrkevich PHPUnit test was failing (I guess because of env variables that you already updated) so I deployed a new one - https://rundash-staging.codesignal.com/coderunner-test/68.183.124.242/test-results

Albinutte (changes requested):
I love the idea of using a package manager! Very nice 🎉 

A few high-level comments:

1. Image is now 1 GB larger:
    image

   Is it expected? @Anastasiya-Zhyrkevich do you know which packages are so big? Are you sure there are no artifacts in the image?
2. I pulled the image locally, and launched it. Looks like `phpunit` is not available there: I get `bash: phpunit: command not found` when I try to launch `phpunit` in there, and the tests in the link @ArDrambyan posted are still failing.
3. Since we're switching to package manager, can we install other packages with package manager as well for consistency? I'm talking about `curl`, `mysql`, `xml` and `mbstring`.

ArDrambyan:
I deployed a couple of more coderunners but `phpunit` test is not passing yet.

ArDrambyan:
> 1. Image is now 1 GB larger:
@Albinutte - I think we should compare `:dev` with `:v8.1.1` because `:new` is based on Ubuntu 20.04 while @Anastasiya-Zhyrkevich is deriving from Ubuntu 16.04
Screen Shot 2021-03-04 at 16 56 46

Anastasiya-Zhyrkevich:
@ArDrambyan  `:v8.1.1` is also mine, from this branch, sorry about this mess, I did not know anything about `dev` tag before.

ArDrambyan:
> @ArDrambyan `:v8.1.1` is also mine, from this branch, sorry about this mess, I did not know anything about `dev` tag before.

Ah, I see 😅

ArDrambyan:
Hi, @Anastasiya-Zhyrkevich @Albinutte do we have a reason that this PR is open for such a long time? It's better to avoid keeping PRs open for such a long time because it's becoming hard to maintain them and they can become blockers for future tasks. 
I think it would be better to just close it if we don't need these changes anymore.

PR #318, opened 2021-03-03
go-junit-reort installed, redundant line removed

KARM99:


go/Dockerfile
```diff
@@ -1,7 +1,11 @@
...
-    && add-apt-repository -y ppa:longsleep/golang-backports \
+RUN add-apt-repository -y ppa:longsleep/golang-backports \
     && apt-get update \
     && apt-get install -y --no-install-recommends golang-go \
     && rm -rf /var/lib/apt/lists/*
+
+# installing go-junit-report for parsing test outputs
+RUN go get github.com/jstemmer/go-junit-report
+
+ENV PATH $(go env GOPATH)/bin:$PATH
```
Albinutte, line 11: Hey @KARM99, could you explain why $PATH needs to be updated here?

KARM99: I'll need to run `go-junit-report` in coderunner. I can write the path there when running the command if there's a problem with modifying the PATH here.

Albinutte: In this case I don't think this works. I pulled the image locally, and here's what I got:
image

KARM99:
The message of the commit I wanted to write was `$(go env GOPATH) to literal` :sweat_smile:

Albinutte (approved):
Nice! 👍

PR #325, opened 2021-03-17
Ubuntu version 20.04

### LINKED ISSUE #324: Release first layer of docker images
After updating each docker image one by one we are ready to start drafting releases for each of them, for the beginning let's merge `276-ubuntu-version-20` changes with `master` and draft releases for first layer of images

ArDrambyan:
Closes #324 

## Changes

Resolved conflicts with the master branch


.circleci/config.yml
```diff
@@ -17,7 +17,7 @@ jobs:
             echo "$DOCKERHUB_PASSWORD" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
 
             echo List of updated files
-            FILES=$(git diff-tree --name-only -r --diff-filter=ACMRT origin/master HEAD)
+            FILES=$(git diff-tree --name-only -r --diff-filter=ACMRT origin/276-ubuntu-version-20 HEAD)
```
menewman, line 20: Should we change this back to `origin/master` before merging?

menewman (approved):
Phew, what a big merge! Nice to see it all coming together. Approving this overall since the images should all be working individually, but with one small comment about the CircleCI/Git change detection.

.circleci/config.yml
ArDrambyan: yeah, I forgot to revert it, thanks!

ArDrambyan:
@menewman - CI is failing because it tries to rebuild everything :D

menewman:
@ArDrambyan - If it's just a one time failure due to the way we calculate changed files in CircleCI, I think the answer is probably a GitHub admin override, which I'm happy to help with. But also, the other thing we should do is release these base updates in layers? First, base. Then, update the Dockerfiles to all inherit from base v9.0, then the next layer of dependencies?

ArDrambyan:
> If it's just a one time failure due to the way we calculate changed files in CircleCI, I think the answer is probably a GitHub admin override, which I'm happy to help with

It would be great!





> But also, the other thing we should do is release these base updates in layers? First, base. Then, update the Dockerfiles to all inherit from base v9.0, then the next layer of dependencies?

I think yes, maybe I'll create intermediate branches to test dockerfiles chunk by chunk but the main idea is to do deploy them layers by layers

menewman (changes requested):
Based on our conversation in Slack, I'm setting status to "request changes" to await the strategy of splitting into small PRs that we can merge and release, layer by layer

PR #339, opened 2021-03-24
Use gradle build for java initial build

roma0297:


java/Dockerfile
```diff
@@ -42,15 +33,16 @@ RUN apt-get update \
...
-    JACKSON_VERSION='2.12.1'
+    JACKSON_VERSION='2.12.2' \
+    MYSQL_CONNECTOR_JAVA_VERSION='8.0.23'
 
-# download dependencies from maven repository with build.gradle script
+# Upload build.gradle script
 COPY build.gradle /usr/share/java/build.gradle
+
+# The command below downloads the dependencies and stores them in the machine memory.
+# It should NOT be removed because it will be used to run kotlin code in the coderunner.
```
menewman, line 46: Just wondering -- Is this reference to Kotlin a copy-paste thing, or is it intentional?

java/Dockerfile
```diff
@@ -21,15 +21,6 @@ RUN apt-get update \
     && rm /opt/gradle-6.7-bin.zip \
     && rm -rf /var/lib/apt/lists/*
 
-# install maven
```
menewman, line 24: Might this cause a problem for filesystem tasks using Maven as the build tool?

menewman (commented):
Glad to see this project picking up steam; nice! 🚂 

I can see that the Dockerfile itself builds, but for this particular kind of change it seems really closely related to what we run on the coderunner, so I have to ask -- is there a deployed coderunner that uses this new image that shows everything is working okay?

java/Dockerfile
roma0297: Oh, that's a copy-paste, fixed it :D

menewman: haha, great

roma0297:
@menewman you're right, we'll need to change some things on the coderunner first. As far as I understand I'll be only to do that once we merge this PR and I create a new tag and the docker image emerges in the Dockerhub, right? Or is there a way to do that without actually merging this PR?

java/Dockerfile
menewman: Ah, I see! Are we sure that there are no Maven tasks in our database at all (on prod)? Including ones created by B2B users, not only ones created by our content team?

If so, I think we can have a discussion about whether we want to only support Gradle going forward, but I think it does warrant a discussion of the pros and cons before removing it. Then, to be safe, we'd also want to make some changes on the product side first to remove Maven as an option when creating a filesystem task through the UI:
Screen Shot 2021-03-23 at 6 14 15 PM

Otherwise, it will be possible to build a task that does not actually run on our newest infrastructure.

menewman:
> you're right, we'll need to change some things on the coderunner first. As far as I understand I'll be only to do that once we merge this PR and I create a new tag and the docker image emerges in the Dockerhub, right? Or is there a way to do that without actually merging this PR?

Yes, there actually is a way to test before this PR is merged! One way is to use the `dev` tag for this Docker image (CircleCI automatically pushes to the `dev` tag when a PR is created in this repo). However, there might be some conflicts since @ArDrambyan is making a bunch of updates in this repo as part of another project. As a result, I've gone ahead and created two new build rules in DockerHub that will automatically build Docker images to the tags `kotlin-gradle` and `java-gradle` when you push to these two branches.

So, you can test on the coderunner by updating the image tags/versions from `v8.1.0` (or whatever they actually are; didn't check) to `kotlin-gradle` and `java-gradle`.

java/Dockerfile
roma0297: As discussed in our chat in Slack, it makes sense to revert this change and add the maven support back 👍 It's already done

DimaPhil (approved):

Albinutte:
I'm going to unassign myself, since Michael beat me to reviewing it 🙂

roma0297:
Hi @menewman, you can take a look at the changes made to the coderunner in another PR: https://github.com/CodeSignal/coderunner/pull/1114
There is also a deployed instance of this new coderunner: http://67.205.158.184:12345/

menewman (changes requested):
Thanks for the coderunner PR, Roman!

Although the changes here seem okay in their own right, setting status to request changes to reflect that the coderunner-side tests seem to be failing on deployed coderunners for this branch:
Screen Shot 2021-03-25 at 1 42 01 PM

ArDrambyan:
Hi @menewman @roma0297 do we need this and a couple of other PRs? I see 4 PRs that were opened more than 6 months ago, I think it will be good to finalize and merge or close them if we don't these features anymore

DimaPhil:
I guess we still need them, but after talk with @menewman we decided that product team will take care of finalizing them when they have time, so leaving this question to Michael :)

menewman:
@ajenks-3MK29A picked this project up and is working on it in separate PRs. @ajenks-3MK29A, are these PRs helpful to keep open for your reference, or are they okay to close?

ajenks-3MK29A:
> are these PRs helpful to keep open for your reference, or are they okay to close?

They're safe to close, the parts I needed from it have been borrowed already. Great work getting it this far!

PR #340, opened 2021-03-24
Make sure that kotlin-scripting-compiler-embeddable is loaded during …

roma0297:
The idea behind this commit was to add `kotlin-scripting-compiler-embeddable` to the list of explicit dependencies so that is loaded during the build phase and caching is applied thus enabling us to run the `gradle run` in offline mode with `--offline` flag

### DIFF
--- a/kotlin/build.gradle
+++ b/kotlin/build.gradle
@@ -27,3 +27,7 @@ sourceSets {
 application {
     mainClassName = 'MainKt'
 }
+
+compileKotlin {
+    kotlinOptions.allWarningsAsErrors = false
+}
--- a/kotlin/build.gradle
+++ b/kotlin/build.gradle
@@ -16,7 +16,7 @@ dependencies {
     implementation "org.jetbrains.kotlin:kotlin-scripting-compiler-embeddable"
     implementation "org.jetbrains.kotlin:kotlin-serialization"
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8"
-    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core"
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.4.2"
     implementation "org.jetbrains.kotlinx:kotlinx-serialization-json:1.1.0"
 }
 
--- a/kotlin/build.gradle
+++ b/kotlin/build.gradle
@@ -1,5 +1,5 @@
 plugins {
-    id 'org.jetbrains.kotlin.jvm' version '1.4.10'
+    id 'org.jetbrains.kotlin.jvm' version '1.5.0-M1'
     id 'application'
 }
 
@@ -8,10 +8,12 @@ repositories {
 }
 
 dependencies {
-    implementation fileTree(dir: "/usr/share/java", include: "*.jar")
-    implementation platform("org.jetbrains.kotlin:kotlin-bom")
-    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8"
+    implementation "com.fasterxml.jackson.dataformat:jackson-dataformat-csv:$System.env.JACKSON_VERSION"
+    implementation "com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:$System.env.JACKSON_VERSION"
     implementation "com.fasterxml.jackson.module:jackson-module-kotlin:$System.env.JACKSON_VERSION"
+    implementation "mysql:mysql-connector-java:$System.env.MYSQL_CONNECTOR_JAVA_VERSION"
+    implementation "org.jetbrains.kotlin:kotlin-scripting-compiler-embeddable"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8"
 }
 
 sourceSets {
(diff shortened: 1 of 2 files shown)


kotlin/build.gradle
```diff
@@ -8,10 +8,12 @@ repositories {
 }
 
 dependencies {
-    implementation fileTree(dir: "/usr/share/java", include: "*.jar")
-    implementation platform("org.jetbrains.kotlin:kotlin-bom")
-    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8"
+    implementation "com.fasterxml.jackson.dataformat:jackson-dataformat-csv:$System.env.JACKSON_VERSION"
+    implementation "com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:$System.env.JACKSON_VERSION"
     implementation "com.fasterxml.jackson.module:jackson-module-kotlin:$System.env.JACKSON_VERSION"
+    implementation "mysql:mysql-connector-java:$System.env.MYSQL_CONNECTOR_JAVA_VERSION"
```
menewman, line 14: Hm, it's not really anything "wrong" with this PR, but this might not work until we have a chance to release the prior change (since this env var might not be defined, if it was just added)

kotlin/build.gradle
```diff
@@ -1,5 +1,5 @@
 plugins {
-    id 'org.jetbrains.kotlin.jvm' version '1.4.10'
+    id 'org.jetbrains.kotlin.jvm' version '1.5.0-M1'
```
menewman, line 2: Hm, I think updating the Kotlin version is great, but I'd probably feel safer sticking to the latest released version (1.4.31) rather than jumping into the preview version

kotlin/build.gradle
```diff
@@ -8,10 +8,12 @@ repositories {
 }
 
 dependencies {
```
menewman, line 10: If you're willing to also add `kotlinx.serialization` and `kotlinx.coroutines` it would be much appreciated!
https://kotlinlang.org/docs/serialization.html#example-json-serialization
https://github.com/Kotlin/kotlinx.coroutines/blob/master/README.md#gradle

It looks like we're manually including the coroutines plugin when compiling in the coderunner using `kotlinc`, so we'd want to adjust that to be available through the Gradle file if we change the way we run Kotlin tasks, I think.

Serialization is a new feature so I wouldn't block this PR on that; would just be nice so that we could easily do some JSON-parsing in the language template (for example, for the string input-parsing project -- see CodeSignal/coderunner#1031)

menewman (commented):
Yay, I'm excited to see this one too! 🙌 

Left a couple of small comments, and I have the same high-level comment that I put on the Java PR:
> I can see that the Dockerfile itself builds, but for this particular kind of change it seems really closely related to what we run on the coderunner, so I have to ask -- is there a deployed coderunner that uses this new image that shows everything is working okay?

kotlin/build.gradle
DimaPhil: Agree with Michael, let's stick with the stable version

ArDrambyan:
Hi all :) As you already know, right now I'm bumping up the Ubuntu version for all dockerfiles and I'm afraid that my changes will result to merge conflicts. I'm sorry about that 😅

kotlin/build.gradle
roma0297: True, we decided to introduce this thing to keep the versions in sync, the order of deployments is important

kotlin/build.gradle
roma0297: It's just a couple of extra lines, I'm gonna add these dependencies :)

kotlin/build.gradle
roma0297: I was under the impression that M1 is also stable, it's just a milestone but I can roll back a bit

kotlin/build.gradle
roma0297: Done 👍

menewman: Awesome 🙌

menewman (commented):
Thank you for the updates! The code changes in this PR look good 👍 

I think I can't approve in good faith until after approving the Java one (since this builds on that), so I guess I'd better go give it another look first

roma0297:
Hey @menewman, I cannot see `kotlin-gradle` tag which you created yesterday in the dockerhub, could you add it again please? I already created a PR to the coderunner repo for Kotlin but it won't work without appropriate docker image: https://github.com/CodeSignal/coderunner/pull/1115

menewman:
@roma0297 - The build rule is in place, but DockerHub currently cannot successfully build this Kotlin Dockerfile as-is because of the issue described here:
https://github.com/CodeSignal/dockerfiles/pull/340#discussion_r600077714

For example, here's the most recent failed build attempt:
https://hub.docker.com/repository/registry-1.docker.io/codesignal/kotlin/builds/260d36c9-cd2b-4aa7-8148-6e6d8b7d2e1c

If you want to try without waiting on Java, some options: we could temporarily hardcode the MySQL connector version, _or_ we could make this inherit from...
```
FROM codesignal/java:java-gradle
```

...instead of `codesignal/java:v8.1.2`.

menewman (changes requested):
Setting status to request changes just to reflect that my status is waiting on testability

PR #348, opened 2021-04-07
Add a groovy script to compile and run scala code

roma0297:


menewman (commented):
Changes make sense! The build is failing because the Java version needs to be released first, but I guess that's not really surprising

Holding on approving since this isn't quite ready to merge

menewman:
Nice to see the Scala version bump sneaking in as well 😄

PR #349, opened 2021-04-07
Add a groovy script to compile and run groovy code

roma0297:


menewman (commented):
Mostly copying review from the other PR:

Changes make sense! The build is failing because the Java version needs to be released first, but I guess that's not really surprising

Holding on approving since this isn't quite ready to merge, although I don't really have specific change requests (just need to see passing coderunner on dev images, then release Java, then release these dependent images)

PR #352, opened 2021-04-29
[Ts] Add @types/mongoose node module

Anastasiya-Zhyrkevich:


Albinutte (commented):
Did you mean to publish a new JS image, and update the base image in this MR? 

You should also update base image of `coffeescript` then, since it's also built on top of JS.

typescript/Dockerfile
```diff
@@ -8,5 +8,5 @@ RUN npm install -g \
     @types/chai \ 
     @types/express \
     @types/node \
-    @types/mocha 
-    
+    @types/mocha \
+    @types/mongoose
```
Albinutte, line 12: There are trailing whitespaces in here:
```suggestion
    @types/mongoose
```

Albinutte (approved):

PR #355, opened 2021-05-04
Update Java to 16, Gradle to 7.0, Maven to 3.8.1

ajenks-3MK29A:
Part of [#1160](https://github.com/CodeSignal/coderunner/issues/1160)

## Changes
* Updated Java from v15 to v16
* Updated Gradle from 6.7 to 7.0 for Java 16 compatibility
* Updated Maven from 3.6.3 to 3.8.1

OpenJFX is still at latest, 11.0.2

## Testing
Verify the CircleCI build. It may be beneficial to test a Java 16 language feature like record classes.


java/Dockerfile
```diff
@@ -2,32 +2,32 @@ FROM codesignal/ubuntu-base:v9.0.0
...
         libmysql-java \
         maven \
     && rm -rf /var/lib/apt/lists/*
 
 # install gradle
 RUN apt-get update \
     && apt-get install -y --no-install-recommends unzip \
-    && wget https://services.gradle.org/distributions/gradle-6.7-bin.zip -P /opt/ \
-    && unzip /opt/gradle-6.7-bin.zip -d /opt/gradle \
+    && wget https://services.gradle.org/distributions/gradle-7.0-bin.zip -P /opt/ \
```
menewman, line 16: Since this `7.0` is repeated a few times, it might be nice to extract a Gradle version ENV variable so that it's easier to update next time? I won't block on it, though.

Same thing for the Maven version below and _arguably_ the Java version, but that one is a little more finicky

menewman (approved):
This looks good to me! If you want to release a new Java image before I get to releasing the first-layer dependencies of `ubuntu-base`, that's okay -- I recommend releasing at version `v9.1.0` in that case.

Left a minor comment about a pre-existing thing

java/Dockerfile
ajenks-3MK29A: I moved everything that is versionable into the top, collated and ready to rock!

menewman (approved):
Thank you for cleaning that up! This should be even easier to update in the future now 👍

PR #353, opened 2021-05-04
Update NVM, Node, Mongo versions

menewman:
No associated issue. Just wanted to update a couple of quick versions while we're working on version updates this week.

## Changes
**In base image:**
- Updated NVM version from 0.35.3 to 0.38
- Updated Node version from 14.14.0 to 14.16.1 (most recent LTS version)
- Updated MongoDB version from 4.4.3 to 4.4.5

## Testing
Verify that the CircleCI status check is able to build the image correctly.

PR #356, opened 2021-05-06
Update first-level dependencies of base, + Python 3, Swift minor versions

menewman:
**Changes:**
Follow-up to [this previous PR](https://github.com/CodeSignal/dockerfiles/pull/353) that updated the base image to bump NVM, Node, and Mongo. Now that the base image is updated and released at `v9.2.0`, we should update all direct dependencies of the base image.

While I was at it, I also bumped the minor versions of Python 3 and Swift, which are out of date but should be safe to upgrade in-place.

**Testing:**
For now, just verify that the CircleCI build step passes!


ajenks-3MK29A (approved):
So exciting! Combining this with the Ruby and Java-derived updates after this constitutes a big lift for our users.

PR #358, opened 2021-05-07
Updated GCC version to 11+

### LINKED ISSUE #357: Update GCC version to 11+
In the scope of this issue lets update the `GCC` version to 11+ and also prebuild `testsRunner.o` object file which will be needed for C++ unit tests, for more details see https://github.com/CodeSignal/coderunner/pull/1176#issuecomment-833670546

ArDrambyan:
Closes #357 

- Updated GCC version to 11.1.0
- Added `testsRunner.o` object file which should significantly decrease compilation time for the `C++` unit tests


Albinutte (approved):

menewman:
@ArDrambyan - Hi, just wanted to ping on the release version for this change. Looks like this PR was released with v9.0.1? The latest build for C++ is actually v9.2.0, although we haven't yet updated the coderunners to use the latest version. As a result, it makes sense to release C++ at v9.2.1 or greater, I think

PR #362, opened 2021-05-10
Update Java derivatives to Java base image v9.2.0

ajenks-3MK29A:
Closes #1160

## Changes
* Update Clojure, Groovy, JavaProject, Kotlin, and Scala images to use Java base image v9.2.0
* Reorganized Clojure Dockerfile to separate versioning variable
* Reorganized Kotlin Dockerfile and build.gradle to separate versioning to variable

## Testing
Verify CircleCI build.


ajenks-3MK29A:
`java-project` is the only hold out. It doesn't like the Maven update, and there are some clues as to why, but no concrete answers. Will investigate a bit more in the morning.

kotlin/build.gradle
```diff
@@ -20,4 +20,15 @@ sourceSets {
 
 application {
     mainClassName = 'MainKt'
-}
\ No newline at end of file
+}
+
+// Gradle 7.0 side effect - https://youtrack.jetbrains.com/issue/KT-46165
```
menewman, line 25: Worth noting -- they claim this is fixed in 1.5.0:
https://youtrack.jetbrains.com/issue/KT-46165#focus=Comments-27-4845442.0-0

Might be worth adding to the comment so that we know it's safe to remove this whole block on upgrade?

kotlin/Dockerfile
```diff
@@ -1,4 +1,6 @@
-FROM codesignal/java:v9.0.0
+FROM codesignal/java:v9.2.0
+
+ENV ORG_GRADLE_PROJECT_kotlinVersion=1.4.10
```
menewman, line 3: Wow, didn't know about this magical environment variable syntax for Gradle

menewman (approved):
:shipit: lgtm!

ArDrambyan:
Hi @menewman @ajenks-3MK29A, I need to ask you to exclude `Kotlin` changes before merging, current Kotlin is not compatible with coderunners, so we are using `v8.0.2` yet, which is too old and doesn't have terminal support, which is really needed for the android interview, so I'm going to roll back some `Kotlin` changes to be able to create a working image with latest `ubuntu-base` changes

menewman:
Thanks for fixing the Kotlin issue, Aram! Since you already updated Kotlin in your change to be based on Java v9.2.0, I think it's now unnecessary for Alex to make further changes in this PR aside from resolving merge conflicts.

ajenks-3MK29A:
@menewman @ArDrambyan I've incorporated Aram's changes while still maintaining the change that allows controlling the Kotlin version in one place (the Dockerfile). Does this look okay? I'm about to spin up a test image to make sure the tests pass with all of these changes.

menewman (approved):
I think these latest changes should be safe, yeah!

PR #368, opened 2021-05-14
Update $PATH in Java so Maven installs correctly

ajenks-3MK29A:
Before this update, the `M2_HOME` variable wouldn't replace correctly in the `PATH`, leading to it defaulting to the `mvn` binary in `/usr/bin`. Not 100% sure why it wasn't replacing correctly, but this fixes it.

Previously this wasn't really an issue, but now that Maven has been updated past what is auto-installed, it has caused friction, especially with `java-project`, where it will error out because Maven 3.6.x isn't compatible with Java 16.

Once this is merged, I can update `java-project` and get everything pushed up to Docker to put https://github.com/CodeSignal/coderunner/issues/1160 to bed.


ajenks-3MK29A:
Forgot to pull latest from `master` before creating the PR, ignore the "11 commits", there's only the one lol.

menewman (approved):
That's very subtle!

PR #371, opened 2021-05-20
Small naming adjustment to match Google coding style

ArDrambyan:
Just renamed `testsRunner` to `tests_runner`

image is pushed with tag `codesignal/cpp:v9.2.1` to avoid changes in coderunners


Albinutte (approved):
> image is pushed with tag codesignal/cpp:v9.2.1 to avoid changes in coderunners

Oooh, that's why https://github.com/CodeSignal/coderunner/issues/1216 happened! Sorry, I should've realized that there's no way to avoid making changes on the coderunner side 😞 

Let's get it all merged 'n' fixed quickly 🚀

PR #384, opened 2021-06-15
Install Go packages in module-less mode

Albinutte:
Main part of https://github.com/CodeSignal/coderunner/issues/1238. Makes sure the modules are installed globally in a module-less manner.

Please check out the linked issue for instructions on how to test the changes.


menewman (commented):
Thanks for the additional research, and it's great to hear that there might be a simple way to accomplish this without modules!

One thing we should double-check: Will this mess up our ability to execute Go projects with unit tests for filesystem tasks? If so, perhaps we need to make some updates to make sure both kinds of tasks can be executed successfully (either splitting the images or setting this environment variable on/off from the coderunner).

Albinutte:
Good point @menewman! It looks like reverting to module-less structure does break unit testing when local files are being imported. The fix turned out to be pretty simple though: setting `GO111MODULE` to `auto` makes sure that:
1. Globally installed packages are still installed with the `src` directory.
2. Projects with `go.mod` get run in the "module" mode.
3. Standalone files get run in the "module-less" mode.

menewman (approved):
🙌

PR #388, opened 2021-06-22
Database Pool: Update images dependent on ubuntu-base and remove no longer used images

mpowaga:
Summary of changes:

- Updated ubuntu-base/9.2.x to ubuntu-base/9.3.0
- Removed mysql image
- Removed mssql image
- Updated readme.md


elixir/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM codesignal/erlang:v9.2.0
+FROM codesignal/erlang:v9.3.0
```
salbert96, line 1: For the first layer we should update only the images that are based on `codesignal/ubuntu-base`
We don't have `codesignal/erlang:v9.3.0` yet, we need to update that layer after merging this and creating a release.

coffeescript/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM codesignal/js:v9.2.1
+FROM codesignal/js:v9.3.0
```
salbert96, line 1: Same problem here.

salbert96 (changes requested):
@mpowaga I did not write in every place, but we should only update the first layer that is based on ubuntu-base (that's the reason why the ci fails).

erlang/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM codesignal/ubuntu-base:v9.3.0
```
salbert96: I thinks this should be updated

mpowaga: Good catch!

salbert96:
@mpowaga seems like Julia does not have the last version, is it intentional?

mpowaga:
> seems like Julia does not have the last version, is it intentional?

@salbert96 It fails to build. I will update it separately to investigate why it fails.

mpowaga:
I'll create another PR that will update second layer images after new images are pushed to docker hub.

salbert96:
> It fails to build. I will update it separately to investigate why it fails.

👍  we don't have to do it soon, feel free to create an issue and it can be done later, after the database-pool, to not block us.

salbert96 (approved):
🚢

coffeescript/Dockerfile
salbert96: @mpowaga seems like this was updated from `js`. Not a big deal, but we should make sure we build everything that is based on js.

PR #389, opened 2021-06-23
Database Pool: Update second layer images that depend on ubuntu-base

mpowaga:
After this PR gets merged we need to create following releases:

- elixir/v9.3.0
- opencv/v9.3.0
- typescript/v9.3.0
- based-on/java/v9.3.0


salbert96:
Seems like we need to update `clojure`, `coffeescript` too.

salbert96:
@mpowaga it's better to create releases based on `java`, `js`, `erlang`, `cpp` for example, to make sure `coffeescript` is also updated.

salbert96 (approved):
👍

PR #392, opened 2021-06-30
Cleanup Java image, remove double-install of Maven, reduce per-step actions by reusing unzip

ajenks-3MK29A:
# Changes
* Remove double-install of Maven (apt and `wget` binaries), defer to binaries
* Standardize `MAVEN_HOME` references
* Standardize JavaFX path to `/opt/javafx` to avoid hard-coding versions elsewhere
* Reuse `unzip` installation across build steps, cleanup at the end

This will also be used as a test for [REDACTED]


ajenks-3MK29A:
I need to stop reusing branches, Github doesn't seem to  like it at all.

menewman (approved):
Looks solid to me! 👍 

One small comment: This Dockerfile linter produces one warning on the updated version of the file:
https://www.fromlatest.io/#/

> **Line 19: apt-get update with matching cache rm (Optimization)**
> Use of apt-get update should be paired with rm -rf /var/lib/apt/lists/* in the same layer.

Not really a blocker, but what do you think?

ajenks-3MK29A:
> apt-get update with matching cache rm

That's handled [at line 53](https://github.com/CodeSignal/dockerfiles/pull/392/files#diff-2c3cd63eeda6d94e4ac3cc5a0493362bc97cd5f8fa694aea346ee2baf6cec5cbL53). Previously we were installing, uninstalling, and purging the `unzip` package several times. This just sets it up once, goes through the process, then purges it when we're done.

menewman (approved):
🚢  Perfect, ty

PR #399, opened 2021-07-20
Add missing MySQL privilege and binary logging

### LINKED ISSUE #398: Some database tasks are failing because of missing SUPER privilege and binary logging
@lemur99 reports:

> Hello everyone!! Some of database tasks are not working correctly on staging now, but in prod it's okay. You can see it here in picture. I googled the following error: ERROR 1419 (HY000): You do not have the SUPER privilege and binary logging is enabled (you *might* want to use the less safe log_bin_trust_function_creators variable)  and I think the solution is written here: https://stackoverflow.com/questions/56389698/why-super-privileges-are-disabled-when-binary-logging-option-is-on
I think this line of code will help fixing this bug SET GLOBAL log_bin_trust_function_creators = 1; 

screen_shot_2021-07-19_at_17 20 45

mpowaga:
Closes #398.


salbert96 (approved):
👍  this only requires updating the database pool image, right?

mpowaga:
Yes, what version number should we use? @salbert96

salbert96:
`v9.4.1` ? since this is a bug-fix

PR #401, opened 2021-07-22
Update list of frequently used dependencies

roma0297:
I did a small analysis of all the Java/Kotlin filesystem tasks which are currently on PROD and I got the following statistics:

```
org.junit.jupiter:junit-jupiter-api:5.6.2 — 38
org.junit.jupiter:junit-jupiter-engine:5.6.2 — 38
org.junit.jupiter:junit-jupiter-params:5.6.2 — 19
junit:junit:4.12 — 20 + 16
org.jetbrains.kotlin:kotlin-stdlib-jdk8 — 9
org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.8 — 7
org.hamcrest:hamcrest-library:2.1 — 6
org.mockito:mockito-core:2.1.0 — 6 + 2
org.springframework.boot:spring-boot-starter-web — 3
com.fasterxml.jackson.core:jackson-databind:2.0.1 — 2
...
```

I added these most frequently used deps to Dockerfile and removed ones which are not used often. Also I'm gonna make sure that we use precisely these versions for all our tasks in `content` project. Also I found out that `maven-compiler-plugin` must be of the same version as Maven itself otherwise this plugin will be downloaded again.


menewman (approved):
@roma0297 - Approving these code changes! Thanks for doing the research.

That said, I think the important part of this PR isn't the actual code change, but probably the research/analysis that informed which libraries to keep and which to remove. 🙂 

For that part, I'm taking your word for it since it's not really visible to me here. But, the likely worst-case impact is just that some different Java filesystem tasks might have slower startup costs when running

menewman:
@roma0297 - I do wonder if maybe it's worth weighting the number of run requests or submissions tied to a task in order to figure out which are the most important, compared to just number of tasks. For example, 1 library used in the Java SCA or in a Java test template would be disproportionately important compared to other Java tasks that aren't used.

DimaPhil (approved):

PR #402, opened 2021-07-28
Update Junit dependency and add missing embeddable Kotlin compiler

### LINKED ISSUE #400: Update java-project Dockerfile to make gradle and maven builds faster
The idea is to make a review of most frequently used dependencies and pre-download them to reduce build time

roma0297:
Update Junit version because latest release enables us to execute test classes in specific order: https://junit.org/junit5/docs/5.8.0-M1/release-notes/ and also add kotlin compiler for kotlin filesystem tasks

closes #400


DimaPhil (approved):

java-project/Dockerfile
```diff
@@ -6,9 +6,10 @@ FROM codesignal/java:v9.4.0
 # -Dartifact=groupId:artifactId:version[:packaging][:classifier]
 #   specifies the target artifact. Packaging is default `jar` but might be `pom`
 RUN mvn dependency:get -Dartifact=junit:junit:4.12 \
-    && mvn dependency:get -Dartifact=org.junit.jupiter:junit-jupiter-engine:5.6.2 \
-    && mvn dependency:get -Dartifact=org.junit.jupiter:junit-jupiter-api:5.6.2 \
-    && mvn dependency:get -Dartifact=org.junit.jupiter:junit-jupiter-params:5.6.2 \
+    && mvn dependency:get -Dartifact=org.junit.jupiter:junit-jupiter-engine:5.8.0-M1 \
```
menewman, line 9: Just want to call this out -- `M1` implies that this is a milestone preview of the upcoming release, not a stable release.

https://commons.apache.org/releases/versioning.html#Milestone_Releases
> While every effort should be made to ensure the quality of released code, "milestone" releases are essentially provided as-is with **no guarantees of stability or maintenance**.

(Emphasis mine.)

If this is for content we're building and we feel good about the milestone, I guess I have no objection to making the intentional choice to go with a milestone version, but if it could affect customer-created content it would be good to be careful about it

menewman (commented):
Exciting that this is now possible in JUnit!

java-project/Dockerfile
DimaPhil: @menewman Thank you for the context! We know that's a milestone and possibly not stable release, but it turns out that the feature we really need for ICA (ordering tests in the report for Java) is properly available only in this release, and we didn't find a way to make it work on earlier releases. We've been also thinking about keeping both versions in the container, but for the content hosted on GitHub we should be able to resolve all issues if there are any, and not having other versions in cache just increases build time by 2-3 seconds (and we anyway can't guess the version the client would like to use)

menewman: As long as you've considered it, okay! Makes sense 👍

menewman (approved):

PR #403, opened 2021-08-02
.Net Core / Reduce build log output, improve build time

vitalik239:
Closes https://github.com/CodeSignal/coderunner/issues/1396
Closes https://github.com/CodeSignal/content/issues/6827

* .Net Core update to LTS 3.1
* Removed separate Asp.Net installation

All tasks will be migrated from Asp.Net 2.2 the same way as https://github.com/CodeSignal/coderunner/pull/1398
Installing separete run-time takes about 4 seconds during each request, so worth migrating the content.


dotnetcore/Dockerfile
```diff
@@ -15,10 +15,16 @@ ENV DEBIAN_FRONTEND='noninteractive'
 # Install .NET SDK
 RUN apt-get -yq install apt-transport-https --no-install-recommends \
   && apt-get update \
-  && apt-get -yq install dotnet-sdk-3.0 --no-install-recommends \
-  && apt-get -yq install aspnetcore-runtime-2.2 --no-install-recommends \
```
menewman, line 19: > Installing separete run-time takes about 4 seconds during each request, so worth migrating the content.

If pre-installing the separate runtimes gives a speed boost to older content that's still running on the older runtimes, is there a reason to remove them beyond clutter cleanup? I would feel safer about this if we knew for sure that no current tasks are running on 2.2 or 3.0 before we release this new Docker image 🤔

menewman (commented):
Great initiative, Vitaliy! Not requesting a change, but leaving a question about our upgrade strategy

dotnetcore/Dockerfile
vitalik239: 1. There are only 3 tasks dependant on Asp.Net 2.2, they can and should be migrated before the image release.
2. The ones with 3.0 target are safe, because it is compatable with 3.1.

This cleanup is an attempt to improve ICA experience. 25% boost for each run is very attractive.

menewman: > There are only 3 tasks dependant on Asp.Net 2.2, they can and should be migrated before the image release.

That's good news! Can we go ahead and migrate them to 3.0 now before merging this PR? It will make me feel better about the overall safety of the change 😄 

Also, can you confirm whether this also applies to customer-created tasks, or just ones in the content repo?

vitalik239: I've created a PR with content tasks migration to .Net 3.0 https://github.com/CodeSignal/content/pull/6870
The only task not in our repo, that relied on old version was [nWaysExerciseCs](https://app.codesignal.com/admin-task/u57ExhkWfDgA5vWJm), I updated the target framework via UI.

Task transition further to 3.1 will be smooth, because no code changes are required.

menewman (approved):
Great, thanks for confirming the migration!

PR #404, opened 2021-08-11
Install minimal ruby on rails dependencies

salbert96:
I did not check this docker image yet, but when setting up from the terminal this is enough for a minimalistic application

Screen Shot 2021-08-11 at 23 46 26

I intentionally fixed the versions since sometimes different sqlite versions don't work with rails.


salbert96:
@menewman  I did not validate the docker image yet, but I think this should be enough. Let me know if you think it's okay to go with this approach.

ruby/Dockerfile
```diff
@@ -47,5 +47,12 @@ RUN gem install \
     rest-client \
     rspec
 
+# Install ruby on rails dependencies
+RUN apt-get -qq update \
+    && apt install --no-install-recommends -y sqlite3 libsqlite3-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN gem install 'rails:6.1.4' 'sqlite3:1.4.2' 'listen:3.6.0'
```
menewman, line 55: Can these be combined with the APT install and gem install commands above for simplicity, or is it important for them to be installed only after everything else?

menewman: Also, do we want to pin the gem versions for these? We usually don't do that for better or for worse, to allow for graceful auto-updating of installed packages (as you can see with the gems above)

salbert96: The order is not important, just combined them to make it easier for the future to separate. But combined it for now.
The reason I fixed the packages was to because the tasks are going to have fixed versions in their `Gemfile`. But let's go with the not-fixed version for now (since I need to deploy the coderunners in a couple of ours).
We can later brainstorm how we can make sure the versions don't break the tasks.

menewman: > the tasks are going to have fixed versions in their Gemfile

Ooh, I see. That's dangerous! Hopefully there's a way around it. Something kind of similar happened with free-coding Java tasks just yesterday, actually, when we pinned a library version that later got bumped in the Docker image:
https://codesignal.slack.com/archives/CDYDBJR4J/p1628726382054200

ruby/Dockerfile
```diff
@@ -45,14 +47,10 @@ RUN gem install \
...
     rest-client \
-    rspec
-
-# Install ruby on rails dependencies
-RUN apt-get -qq update \
-    && apt install --no-install-recommends -y sqlite3 libsqlite3-dev \
-    && rm -rf /var/lib/apt/lists/*
-
-RUN gem install 'rails:6.1.4' 'sqlite3:1.4.2' 'listen:3.6.0'
+    rspec \
```
menewman, line 50: I think these used to be in alphabetic order, but I guess it doesn't really matter 😄

PR #405, opened 2021-08-24
Fixing Julia building issue

### LINKED ISSUE #390: Update Julia image to 9.3.0+ to support Database Pool
The build was failing so we decided to do it later, separately.
See this discussion
https://github.com/CodeSignal/dockerfiles/pull/388#issuecomment-866829399

ArDrambyan:
Closes #390 

## Changes
- Updated Julia's version
- The build was failing on the signature verification, but I don't think we need it so I just removed that part


Albinutte:
> I don't think we need it so I just removed that part

Wellll we don't strictly need it, but it's something nice-to-have ¯\_(ツ)_/¯ The fix appears to be pretty simple: looks like key server got outdated (see [here](https://discourse.julialang.org/t/dockerhub-julia-dockerfile-using-outdated-key-server/64381/2)). Could you check if changing the server like in a commit mentioned in the link will work? If not, we can probably drop it as you suggested.

ArDrambyan:
> Could you check if changing the server like in a commit mentioned in the link will work? If not, we can probably drop it as you suggested.

I tried a couple of key servers and none of them worked for me, but this one is working, thank you!


> UPD: oh, sorry, didn't notice you removed my review request 😅 I'll leave the review to Michal then 👍

I wanted Michal to take a look and confirm that everything else is ready to enable the DB pool of Julia and I thought that the changes are too small for 2 reviewers 😅  But it turned out that requesting your review was correct 😂

Albinutte (approved):
Glad I could help! 😄

PR #407, opened 2021-09-03
Bumping up Dart version

### LINKED ISSUE #406: Upgrade Dart version
Our current `Dart` language version is `1.24` while the latest version is `2.13` and it is nice to keep our language up to date, especially if there is a major version change

ArDrambyan:
Closes #406 

## Changes
Upgraded `Dart` to `v2.13` which is the latest stable version right now


ArDrambyan:
@menewman looks like Dart 2+ is not compatible with Dart 1.x, and additional changes in the `Dart` template are required.  We are okay with that right? The problem is that some user codes will stop working but I think we need to do this some day 😅

menewman:
@ArDrambyan - Yeah, I think it's okay, especially considering how long ago Dart 2 was released. I wouldn't want to break any tasks that are created in Dart, so that would be my main concern if someone has Dart-specific test or interview questions with starter code. 🤔

menewman (approved):

ArDrambyan:
> I wouldn't want to break any tasks that are created in Dart, so that would be my main concern if someone has Dart-specific test or interview questions with starter code.

I'll check tasks on production before merging these changes

PR #415, opened 2021-09-15
PHP composer install

lemur99:
Install composer for `laravel` workspace tasks


RubenAshughyan (changes requested):
Looks good!

Let's add  
`&& rm -rf /var/lib/apt/lists/*` to remove the leftovers after installation.

You can see an example in the same dockerfile.

RubenAshughyan (approved):

Albinutte:
A quick PR stalking note 🥷 : you can use https://www.fromlatest.io/#/ to get recommendations on how to structure dockerfile. For example, it also mentions as a recommendation that `--no-install-recommends` should be used together with `apt-get install` as a tiny optimization. Not a big deal though!

PR #418, opened 2021-09-23
Add Jupyter notebook support for R

salbert96:
Will install python3, pip3, Jupyter notebook and will configure the Jupyter notebook to be used with `R` kernel.

Screen Shot 2021-09-23 at 15 53 17


r/Dockerfile
```diff
@@ -23,10 +23,21 @@ RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 51716619E0
     r-cran-matrixstats \
   && rm -rf /var/lib/apt/lists/*
 
+# Install Jupyter notebook
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends python3-pip   \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN pip3 install notebook
```
salbert96, line 31: @menewman  is it okay that we install python3 this way. We only need it for the Jupiter notebook, so I think we don't need to change the base of this image to python (since the python image has many other packages).

menewman: 🤷 I'm not sure I have an expert opinion on this one -- what are the potential downsides? Just extra maintenance? Seems likely okay to me in that case, assuming this does not drastically increase the size of the image

salbert96: It's just here the python version might be different from the version in the `pythonn3` image. 
But since we only need it to install the Jupiter notebook, I think it should be okay. 
Wanted to make sure it's okay.

menewman (approved):

PR #419, opened 2021-09-25
Update Kotlin build.gradle, copy it to image

ajenks-3MK29A:
## Overview
This PR is one in a series to help the Coderunners reach feature parity between Kotlin and the other languages, chiefly configurable file names, as well as act as a test bed for running JVM languages with Gradle.
The Coderunner has control over environment variables passed into these containers. Changes in another PR allow compilers to customize environment variables right before user code is executed. This environment-controlled `build.gradle` file gives compilers granular control over how Gradle executes.

## Testing
There are no functional changes in this PR (at least for our use cases), so the only testing will be ensuring the checks pass. More PRs will follow updating the other JVM languages to this same standard, but Kotlin is an excellent experiment for this setup because of the level of support and pre-existing infrastructure.


ArDrambyan:
@ajenks-3MK29A changes look good to me, but before merging it would be good to deploy coderunner with these changes and make sure that all tests are passing and we don't have any regression

ajenks-3MK29A:
@ArDrambyan I've updated the PR description to include a regular coderunner link targeting `codesignal/kotlin:dev`

menewman (approved):

PR #424, opened 2021-11-03
Fix MS SQL start script

mpowaga:
[Related issue](https://github.com/CodeSignal/coderunner/issues/1573).

The problem was that `SQLSERVR_STARTED` would only be set to true if `mssql_start.sh` script had to start MS SQL server. If the MS SQL server was already running (for example when we already started the server when creating a database-pool container) it wouldn't be set to true and we wouldn't create a test user which would then make the user's solution fail with a login error.

This issue wasn't easy to notice because only a few run requests failed with this error and the users could re-run the solution in that case. Only after we increased the database-pool queue from 1 to 2 containers did it did become a real problem because MS SQL tests were failing when deploying a new coderunner.

The solution is to remove `SQLSERVR_STARTED` variable. As far as I remember it was introduced to make sure that the setup script is called only once. However, it is safe to run it multiple times because we suppress the error output so if it fails (because the test user already exists for example) it won't stop the execution.


menewman (approved):
Looks like persistent session tests are still going, but I'm very optimistic. Regular tests are passing. Yay!

PR #425, opened 2021-11-05
add dotnetcore into cs image

lizhoucheng:


menewman (changes requested):
@lizhoucheng - Conceptual question -- most of the steps added here are shared with the steps in the `dotnetcore` image, meaning that when we want to make any changes to our .NET Core installation we'd have to maintain them in two places. Would it make more sense just to have one Docker file derive from the other so that the common parts really are shared and not defined twice?

menewman (approved):
Great!

Before merging, can you please also update the README to show the updated dependency hierarchy?

lizhoucheng:
Okay I think the error may not be related to the changes in my PR. I roll back the changes in cs image and it still fails at docker hub. I think I m going to push the changes again and merge them. What do you think @menewman

menewman:
Sounds okay to me

PR #431, opened 2021-11-15
Update Java/Gradle, Add cron to Java, Create system to keep the Gradle daemon running

ajenks-3MK29A:
Sadly a few changes have to be bundled into this one, since Oracle removed their `oracle-java16*` packages.

This PR introduces an entry point script and cron job, which will allow a long-running container to keep the Gradle daemon alive. This should improve the compile time for JVM languages significantly, since the daemon incurs a significant startup cost, anywhere from 5 to 20 seconds.

### `codesignal/java`
- Update Java to v17
- Update Oracle License to v1-3
- Update Gradle to v7.2 (to support Java 17)
- Add `cron` to list of installed packages
- Add a daemon script, entrypoint script and precompiled crontab to the container
- Set `ENTRYPOINT` for the container


menewman (approved):
I don't have any requested changes here, but I'm also not confident that it works either without seeing it in the context of the coderunner -- should probably review & approve both PRs together?

java/Dockerfile
```diff
@@ -16,9 +16,12 @@ RUN wget https://github.com/adoptium/temurin16-binaries/releases/download/jdk-16
 ENV CLASSPATH=/usr/share/java/*:$CLASSPATH \
     PATH=${JAVA_HOME}/bin:$PATH
 
-# Install Java MySQL drivers
-RUN apt-get update \
-    && apt-get install -y --no-install-recommends libmysql-java \
+# Use new MySQL public key, install Java MySQL drivers and cron
```
menewman, line 19: Should this be updated in the base image instead?

java/setup.sh
```diff
@@ -0,0 +1,45 @@
...
+mkdir -p /usercode/FILESYSTEM
+
+cp /usr/share/java/build.gradle /usercode/build.gradle
+cp /usr/share/java/settings.gradle /usercode/settings.gradle
+
+# Even though the file name may change, it won't affect Gradle's build/run time.
+# It just needs some frame of reference as to what it's supposed to build, so
+# that it can optimize the next runs and do it in three seconds instead of 16+
+cat << EOF > /usercode/FILESYSTEM/main.java
+    class main {
```
menewman, line 24: [nitpick] If the file or class name does not matter, maybe we can follow Java convention and uppercase it? Like, `Dummy.java` / `class Dummy`?

menewman (approved):

artahian (approved):

PR #434, opened 2021-12-03
Use Temurin/AdoptOpenJDK for Java

ajenks-3MK29A:
## Overview
Oracle recently released their JDK 17, and due to their release guidelines, they deleted JDK 16 packages from apt.
Because of this, we need to either update to Java 17, or use an alternative build of Java 16, otherwise we cannot build any new releases of the Java container.
Java 17 adoption is a larger project requiring more changes than simply a version bump, therefore this PR is an experiment for using AdoptOpenJDK's builds, called Temurin. They're a very popular group that is very prominent and active in the Java community, so we should be in safe hands in terms of maintainability and security. Recently the project was picked up by the Eclipse Foundation as well, going under the new name Adoptium. While this didn't change overall functionality, this did further cement OpenJDK's position in the Java space as a viable, GPL-2 alternative to Oracle's commercial licensing.

## Impact
There are a couple differences between Oracle JDK and Temurin, but for our use cases none of them apply.
- We don't use Mission Control, the Flight Recorder, or Class Data Sharing.
- The different garbage collector works perfectly well

There are some reports about performance differences or instability due to their biannual release schedule, but they are rare or minimal. For us, neither of these are an issue.


menewman:
@DimaPhil - Heads up about this change! It sounds like it should be a seamless transition, but just wanted you to know

menewman (approved):

DimaPhil:
@menewman Thanks for letting me know! I think this should be a safe transition for our use cases, as OpenJDK support is typically pretty good for any provider, including Temurin

However, transiting to Java 17 which is going to be LTS for a very long time should also make sense, so would be glad to see further updates to it too :) Thank you!

PR #435, opened 2021-12-08
Use New Java v9.5.0 Image in JVM Containers

ajenks-3MK29A:
This should be a seamless transition, but our automated tests will confirm.


menewman (approved):
> The only downside (theoretical but not confirmed) is subsequent apt installs will have an error because of a dependency problem.

Just trying to clarify here -- you mean subsequent install of _any_ package? Or only of the Scala package?

ajenks-3MK29A:
The only thing left to make this work is to install `sbt` and install scala through that. It will avoid dependency problems.
You can do it through aptitude or direct zips from their GitHub

https://www.scala-sbt.org/1.x/docs/Installing-sbt-on-Linux.html
https://github.com/sbt/sbt/releases

PR #436, opened 2021-12-08
Update README.md

lizhoucheng:
update dotnetcore and cs dependency


lizhoucheng:
nvm I can just change in my PR

PR #440, opened 2022-01-10
Bump up PostgreSQL version to v14.0.0

mpowaga:
I added new apt key `apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 467B942D3A79BD29` because MySQL and Microsoft SQL failed to install.


salbert96:
Sorry for the delay, Michal, I will make sure to review it today. This week was too busy since we had new hires.

salbert96 (approved):
LGTM 👍

PR #442, opened 2022-01-26
441: Test updating mysql apt config

miamagana:
This PR fixes build base image error.

MySQL changed the GPG key and therefore our build fails throwing an unsigned download error.


RubenAshughyan (approved):
That looks good 👍

PR #444, opened 2022-01-26
Updated images based on ubuntu base to v9.6.0

miamagana:
This PR updates images that depend on ubuntu base to v9.6


RubenAshughyan (commented):
The changes are technically correct!
Good job @miamagana 

As we talked the only issue is `python2` image build failing. Maybe it is because of some flakiness? You can perhaps rerun the CI flow to check if it'll fail again.

Screen Shot 2022-01-27 at 18 44 42

miamagana:
> The changes are technically correct! Good job @miamagana
> 
> As we talked the only issue is `python2` image build failing. Maybe it is because of some flakiness? You can perhaps rerun the CI flow to check if it'll fail again.
> 
> Screen Shot 2022-01-27 at 18 44 42

Yes! The only thing is that python images fail to build.

I just managed to build a base image locally with the v9.4 and I am now building python2-3 images locally to check if this is happening due to the latest changes or if it is something else. Will conclude the investigation in around an hour.

miamagana:
Finally @RubenAshughyan , python2 and python3 have been removed from this PR due to an issue when building the images that I do not think to have to do with these changes

Albinutte (approved):

PR #448, opened 2022-02-03
php mbstring version needs to be pinned to avoid potential errors

### LINKED ISSUE #449: PHP / Phpunit version mismatch causing trouble
PHP version is 8.0 and the dependencies are the following:

- PHP unit version is 9.x
- mbstring version is latest

One of those dependencies is causing a version mismatch and therefore we have failing tests in _coderunners_ due to a missing required dependency.

miamagana:
Closes #449


lemur99 (approved):
LGTM! 🏅

PR #457, opened 2022-03-16
Replace mvn dependency:get with gradle

roma0297:
The idea of this PR is to replace maven `dependency:get` command with gradle dependencies download. There main reason is that maven `dependency:get` doesn't download transitive dependencies and hence is not so effective. 

In order to prove that this change actually improves the performance I created the coderunner `143.198.166.194` and tried to run problematic tasks on it like `cheapestAirplaneTicketKotlin` or `ridesAggregatorKotlin`. They fail with TLE on the current coderunner and complete successfully on `143.198.166.194`.


java-project/build.gradle
```diff
@@ -0,0 +1,31 @@
...
+
+repositories {
+    mavenCentral()
+}
+
+configurations {
+    toCopy
+}
+
+dependencies {
```
DimaPhil, line 13: Do you think we can make these duplicated library versions variables everywhere? So whenever we have to change the version, it’s just a one-line change

DimaPhil (approved):
Great stuff, and really nice catch about transitive dependencies! Seems to work for me!

Albinutte (approved):
Very nice! 👍

PR #463, opened 2022-05-03
Install dotnet 6.0 version along side 3.1

ericloh:
Added one more apt-get to install version 3.1 as well. Deployed a coderunner to 147.182.143.230 for testing and it can run both 3.1 and 6.0 .NET projects. Just need to change `<TargetFramework>netcoreapp3.1</TargetFramework> `to `<TargetFramework>net6.0</TargetFramework>` to run 6.0 version.


Albinutte (approved):
Great! Looks like the image got almost 200Mb bigger, but I guess this is inevitable.

PR #464, opened 2022-05-05
Update Dockerfile - add priority-queue library to js

malinovsky239:
@ArDrambyan Please take a look.

See https://github.com/CodeSignal/coderunner/issues/1839 for more details.


DimaPhil:
I agree with Ilia that we should limit the list of available data structures, so trie/suffix data structures or anything like that might be too much. However, I feel like we can also add BST and heap for JS, as `std::set` is built-in, and I don't remember having something similar in JS

menewman:
Just wanted to bring up Dima's comment from [here](https://codesignal.slack.com/archives/CFVVDHQ9F/p1649789581710109?thread_ts=1649771677.594029&cid=CFVVDHQ9F): 

Quote:
> Priority queue might be required [....] However, I'm not sure about these two specifically - they don't seem to be too popular (just 175 stars on GitHub). Maybe we should use [more popular ones](https://github.com/montagejs/collections)?

I think it's fair not to include a lib with overly advanced data structures for the reasons you mentioned. But Aram's & Dima's concern about whether the lib is the most widely used & maintained is also a fair one -- worth double-checking what the alternatives are so we pick the one we're least likely to regret! Maybe it does end up being this one in the end, but it's worth researching.

malinovsky239:
* Added a test case verifying that the new library can be imported: https://github.com/CodeSignal/coderunner/compare/js-priority-queue.
* Deployed a coderunner: http://rundash-staging.codesignal.com/coderunner-test/137.184.157.56/test-results.

@ArDrambyan please take another look.

ArDrambyan:
Looks good! 👍 
One small comment related to the new test case:
In the `API/test/raw.js.test.js:324` I see dependency on how `console.log` prints array. I think there is any guarantee that the format will remain the same in upcoming `Node` versions. My suggestion is to change `console.log(numbers)` to `console.log(JSON.stringify(numbers))` because `JSON.stringify` has a strictly defined behavior and we won't need to worry about test case adjustments in the future.

ArDrambyan (approved):

PR #466, opened 2022-05-26
Update Java version -> Java17

ericloh:
Once this is released I will update java-project base image to this released version.

Tests passed: http://rundash-staging.codesignal.com/coderunner-test/167.99.227.237/test-results

Had to update Gradle as well because only 7.3 and above supports java 17: https://docs.gradle.org/current/userguide/compatibility.html


Albinutte (approved):
Great! I ran an existing task with gradle testing tool against the coderunner you deployed and everything seems to be working 👍

PR #469, opened 2022-05-30
update java image version

lizhoucheng:
A follow up for issue: https://github.com/CodeSignal/codesignal/issues/24879

I found that some languages that depend on Java haven't updated. So I update them in this PR

The current version of Groovy does not work with Java 17, so I update it to the latest stable version.

Coderunner PR: https://github.com/CodeSignal/coderunner/pull/1913
Coderunner test: http://rundash-staging.codesignal.com/coderunner-test/165.227.194.83/test-results


Albinutte (approved):
Thank you for noticing this! 👍

PR #471, opened 2022-06-02
[22724] add datadog to our custom meteor cloud image

sig-bo:


sig-bo:
@artahian couldn't get the hostname to change, but if we set the correct tags, it should be enough to work with?

sig-bo:
image

build failed due to a permissions issue i believe

artahian:
> @artahian couldn't get the hostname to change, but if we set the correct tags, it should be enough to work with?

I guess that's a reasonable workaround - tags will likely be our primary indicators anyway especially when the number of instances gets large enough when we wouldn't be able to simply browse them anyway.

sig-bo:
> > @artahian couldn't get the hostname to change, but if we set the correct tags, it should be enough to work with?
> 
> I guess that's a reasonable workaround - tags will likely be our primary indicators anyway especially when the number of instances gets large enough when we wouldn't be able to simply browse them anyway.

Cool. Do you know what's wrong with the build permissions issue?

artahian:
I'm actually not sure why it's failing - double-checked the credentials and they seem to be ok. @menewman do you happen to be familiar with these errors from the CI logs?

> Successfully built 15bd1e312123
Successfully tagged codesignal/meteor-cloud:dev
The push refers to a repository [docker.io/codesignal/meteor-cloud]
>
> 3b89493e: Preparing 
0a814e89: Preparing 
c86bee1e: Preparing 
fba7b5a2: Preparing 
8bdc968e: Preparing 
459170d8: Preparing 
8cae6a05: Preparing 
denied: requested access to the resource is denied

menewman:
I have a theory, yeah. Updated some permissions for this repo in DockerHub; let's see if it helps.

(BTW this is the first PR to update this particular Docker image, so it's the first time we would have potentially run into this CircleCI build issue)

artahian:
Thanks! I didn't realize this was the first update since there were multiple commits in the history, but after looking at it again I realized that they were all made the same day so it was likely all together.

artahian:
That worked! 🎉 
Thanks a lot @menewman 🙌

meteor-cloud/app/setup.sh
```diff
@@ -20,8 +20,12 @@ export PATH="/node-v${NODE_VERSION}-linux-x64/bin:$PATH"
 npm_version=$(/app/select_npm_version.sh)
 npm -g install npm@"${npm_version}"
 
-# NPM install
 pushd programs/server
+
+if [ -f "/app/bundle/programs/server/assets/app/p2d-enabled.galaxy" ]; then
+  export METEOR_SKIP_NPM_REBUILD=1
+fi
```
artahian, line 27: @sig-bo is there a particular reason to add this?

sig-bo: @artahian yep. i updated the existing files under `meteor-cloud` with the latest files from `https://github.com/meteor/galaxy-images/blame/master/galaxy-app/`. the only change is in this file.

artahian (approved):
Looks good! Once merged I will enable this for staging.

PR #473, opened 2022-06-02
Typescript and Coffeescript update

lemur99:
Closes #472 

Since we change, the release of js, I update it in typescript and coffeescript.


salbert96 (approved):
👍

PR #476, opened 2022-06-29
[22644] set env for datadog config file for meteor-cloud custom image

sig-bo:


meteor-cloud/app/run.sh
```diff
@@ -5,13 +5,11 @@ set -e
 # set up datadog config and start it up
 if [[ -n $DD_API_KEY ]]; then
   sh -c "sed 's/api_key:.*/api_key: ${DD_API_KEY}/' /app/datadog.yaml > /etc/datadog-agent/datadog.yaml"
+  sh -c "echo 'env: ${DD_ENV}' >> /etc/datadog-agent/datadog.yaml"
```
artahian, line 8: @sig-bo would this break if DD_ENV is not specified? Do you think it's worth skipping this line in that case to avoid potential syntax issues with an empty `env: ` line?

sig-bo: good point. added in a check.

artahian (approved):

PR #477, opened 2022-07-01
create-candidate-user-in-root-group

dejtor:
only testing :)


typescript/Dockerfile
```diff
@@ -30,3 +30,6 @@ RUN npm install -g \
 # Create a tsconfig.json file
 COPY create_ts_config.sh /typescript/create_ts_config.sh
 RUN bash /typescript/create_ts_config.sh && rm /typescript/create_ts_config.sh
+
+RUN useradd -ms /bin/bash -u 1511 -g 0 candidate && echo "candidate:123455" | chpasswd
+RUN find . -user root -exec chmod g=u {} \; || true
```
salbert96, line 35: I think it's safe to start changing this in ubuntu-base.

typescript/Dockerfile
```diff
@@ -30,3 +30,6 @@ RUN npm install -g \
 # Create a tsconfig.json file
 COPY create_ts_config.sh /typescript/create_ts_config.sh
 RUN bash /typescript/create_ts_config.sh && rm /typescript/create_ts_config.sh
+
+RUN useradd -ms /bin/bash -u 1511 -g 0 candidate && echo "candidate:123455" | chpasswd
```
salbert96, line 34: Should this be `12345`?

salbert96: We don't use the password anywhere, right?

dejtor: > Should this be 12345?

probably yes, will check if there are some restrictions

>We don't use the password anywhere, right?

no, but before I needed it for account creation. Not sure if we need it anymore and will try removing it

PR #478, opened 2022-07-19
Scala update

Albinutte:
Looks like the best way to install Scala these days is with Coursier (docs: [here](https://www.scala-lang.org/download/3.1.3.html)).

Updated Dockerfile to install Coursier and use it to install required `scala` and `scalac` versions.

Please see https://github.com/CodeSignal/coderunner/pull/2003 for the corresponding coderunner changes.


scala/Dockerfile
```diff
@@ -1,11 +1,15 @@
...
-    && dpkg -i scala-${SCALA_VERSION}.deb \
-    && apt-get update \
-    && apt-get install -y scala \
-    && rm -rf scala-${SCALA_VERSION}.deb  /var/lib/apt/lists/*
+    && curl -fL https://github.com/coursier/launchers/raw/master/cs-x86_64-pc-linux.gz | gzip -d > cs  \
+    && chmod +x ./cs \
+    && ./cs setup --yes \
+    && rm -rf ./cs
+
+ENV PATH ${PATH}:/root/.local/share/coursier/bin
```
lizhoucheng, line 11: Why do you need to set this env

lizhoucheng (approved):

scala/Dockerfile
Albinutte: This commands makes sure that `cs` is now a global variable that can be used to install a specific Scala version below.

